PR #38, opened 2026-07-30
Add portal playground mode: anonymous browser-pinned sessions

16francej:
## What

`PORTAL_PLAYGROUND=1` turns a deployment into a public try-it instance with one real admin. An unauthenticated browser navigation (a `GET` that accepts HTML) mints an anonymous `playground-<random>` principal, seals it into the ordinary `portal_session` cookie, and continues. Each visitor's sessions, files, memory, keychain view, and sandbox are pinned to their browser through the same `personal:<principal>` scoping that isolates real teammates — no new isolation machinery.

What playground mode deliberately does **not** change:

- `/auth/login` still runs the full OIDC flow; production still demands the usual OIDC config. That is how the one admin signs in, and admin identity remains the core's `ADMIN_GRANTS`, derived per request.
- `/admin` refuses anonymous sessions outright (HTML card or JSON 403).
- Non-HTML requests without a session still get `401` — the SPA's API calls ride the cookie from the first page load, and bare `curl` never mints.
- Sign-out just clears the cookie; the next visit starts a fresh identity.

## Abuse containment

Minting is rate-limited per client IP (`clientIpOf`, Fly-header aware) through the core's Postgres-backed single-use claim store — the same `/v1/auth/broker/claim` slot pattern the sign-in broker uses — so restarts, blue-green deploys, and multiple portal instances share one budget. If the core can't record the claim, the portal fails closed with a 429 card. Tunables: `PORTAL_PLAYGROUND_MINTS_PER_IP` (default 30) per `PORTAL_PLAYGROUND_MINT_WINDOW_S` (default 3600).

A cleared cookie is a fresh principal, so the per-principal core brakes are soft; the README directs operators to the real ones — `ORG_BUDGET_USD_PER_WINDOW`, `BUDGET_USD_PER_WINDOW`, `RATE_LIMIT_PER_WINDOW`, and pinning a single model via the admin `base-model` / `webui-models` resources — and to run a playground as its own deployment. Known gap, stated in the README: nothing garbage-collects an abandoned visitor's scope yet.

## Moved

`plugins/auth/src/claims.ts` → `plugins/chassis/src/claims.ts` (the claim-store client + `withinRateLimit` slot math), now that both the auth broker and the portal consume it. Chassis is the sanctioned home for plugin↔core plumbing; the log prefix became a parameter.

## New surface

The only new rendered page is the 429 "playground is busy" card, built from the existing portal card system. Rendered from the real `playgroundBusyHtml()` output (the live path needs an exhausted IP budget, so this is a direct render — same bytes):

The playground busy 429 card

## Testing

- New `plugins/portal/test/playground.test.ts` boots the real portal server against a stub core + upstream and covers: mint-on-first-visit with cookie pinning (no re-mint on return), 401 for sessionless API calls, `/admin` refusal for anonymous sessions, `/auth/login` still reaching the IdP, per-IP budget exhaustion → 429 with no cookie, and fail-closed when the claim store refuses.
- Portal: 101/101 tests, typecheck, lint. Auth: 52/52, typecheck (import-path move only).

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/38"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788044677&installation_model_id=19911&pr_number=38&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F38&signature=61aa0ec890578d512657f14de5c165e8b2cb3eb8ef0ca0a1199a95ce72705e88"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/auth/src/index.ts
+++ b/plugins/auth/src/index.ts
@@ -3,7 +3,7 @@ import { pathToFileURL } from "node:url";
 import { json } from "../../chassis/src/http.ts";
 import { portFromEnv } from "../../chassis/src/env.ts";
 import { bootProblems, readConfig } from "./config.ts";
-import { coreClaimStore } from "./claims.ts";
+import { coreClaimStore } from "../../chassis/src/claims.ts";
 import { mailerFor } from "./email.ts";
 import { loadSigningKey } from "./keys.ts";
 import { TokenSigner } from "./tokens.ts";
@@ -27,7 +27,7 @@ export async function startServer(): Promise<void> {
     cfg: CFG,
     signingKey,
     signer: new TokenSigner(CFG.tokenSecret, CFG.issuer),
-    claims: coreClaimStore(CFG.coreApiUrl, CFG.coreSigningSecret),
+    claims: coreClaimStore(CFG.coreApiUrl, CFG.coreSigningSecret, "auth"),
     mailer: mailerFor(CFG),
   });
   const server = createServer((req, res) => {
--- a/plugins/auth/src/server.ts
+++ b/plugins/auth/src/server.ts
@@ -3,7 +3,7 @@ import { readBody, PayloadTooLargeError, serveEmojiFavicon } from "../../chassis
 import { errMessage } from "../../chassis/src/errors.ts";
 import type { AuthConfig } from "./config.ts";
 import { validEmail } from "./config.ts";
-import { claimOnce, withinRateLimit, type ClaimStore } from "./claims.ts";
+import { claimOnce, withinRateLimit, type ClaimStore } from "../../chassis/src/claims.ts";
 import { mintIdToken, pkceMatches, safeEqual, subjectFor, TokenSigner, type AuthRequest } from "./tokens.ts";
 import { ID_TOKEN_ALG, type SigningKey } from "./keys.ts";
 import { renderSignInEmail, type Mailer } from "./email.ts";
--- a/plugins/auth/test/helpers.ts
+++ b/plugins/auth/test/helpers.ts
@@ -1,7 +1,7 @@
 import { createServer, type Server } from "node:http";
 import { createHash, generateKeyPairSync, randomBytes } from "node:crypto";
 import { readConfig, type AuthConfig } from "../src/config.ts";
-import type { ClaimStore } from "../src/claims.ts";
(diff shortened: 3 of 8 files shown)


16francej:
Pushed a hardening commit after two independent adversarial reviews (fresh-context review agent + Codex challenge):

- **Boot refusals** — `PORTAL_PLAYGROUND` now refuses to start with `PORTAL_COOKIE_DOMAIN`, `PORTAL_APPS_DOMAIN`, or `PORTAL_DEPLOYMENTS_ENABLED` (those surfaces never see the `anon` flag and would honor an anonymous cookie), and refuses mint knobs outside the core's real bounds (limit 1–64 claim slots, window 60–86400s) instead of silently 429ing every visitor. An explicit-but-invalid value now fails boot instead of silently becoming the default.
- **IPv6 minting is bucketed per /64**, so a routed prefix can't rotate through fresh budgets; boot warns when the socket address would make everyone behind a reverse proxy share one bucket (`PORTAL_XFF_TRUSTED_HOPS`).
- **Anonymous sessions are refused `/connect/*` and `/drop/*`**, so real OAuth tokens and dropped secrets can't be attached to a throwaway principal that a cleared cookie orphans.
- New tests: anon surviving sliding renewal, connect/drop refusal, the /64 bucketing, and the boot refusal matrix. Portal 105/105.

Documented residual (by design, now stated plainly in the README): the `anon` flag never crosses the portal identity boundary — inside the playground org a visitor is an ordinary internal principal, so a playground must be its own deployment with nothing sensitive at `org:` scope. Also still open, as before: no GC for abandoned visitor scopes.

PR #39, opened 2026-07-31
Document playground deployments in the deployment workflow

16francej:
Follow-up to #38, which added playground mode but documented it only in the portal README — an operator following the deploy path could never discover it.

Adds a "Playground mode" subsection to step 3 of the deployment workflow (`cli/templates/deployment/deployment.md`, materialized into every org's deployment directory by `qm init`), covering:

- the `env.portal.PORTAL_PLAYGROUND` stack snippet
- the its-own-deployment rule: visitors are ordinary internal principals of the org, so nothing sensitive at org scope, no real connector credentials, no company data
- the real brakes to set in the same pass: `ORG_BUDGET_USD_PER_WINDOW` as the hard ceiling, per-principal `BUDGET_USD_PER_WINDOW` / `RATE_LIMIT_PER_WINDOW`, and a pinned model via the Admin page's base-model / web-UI model resources
- the boot refusals and anon restrictions the portal enforces, the scope-GC gap, and a pointer to the portal README for the mechanism

Docs only; CLI init tests pass (17/17).

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/39"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788049131&installation_model_id=19911&pr_number=39&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F39&signature=be8208eae500f63b25a3fa7f510128b4b48368d1fae5a2bd7d4696be7d5afc39"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Why the CLI version bump

`cli/templates/deployment/deployment.md` ships inside the `@yc-software/qm` npm package (it's what `qm init` materializes into a deployment repo). CI's "CLI version bump" check refuses any change to packaged files without a version bump past the published `0.1.2`, so this PR bumps `cli/package.json` to `0.1.3`. Doc template change only — no CLI code changes ride along.

## Also: enforce the web-UI model picker at every core entry

Reviewing the docs claim surfaced a bypass: `webui-models` only gated the explicit per-turn `model:` option. `PUT /v1/runtime-config` accepted any model an approved harness supports, so a user could pin an excluded model as a scope override and turns ran it without re-validation.

Fixed at both entries: the route now refuses a model outside the configured picker (`400 model_not_enabled`), and web turn resolution refuses a resolved scope override outside it (covers overrides that predate a picker change). The org default stays exempt — the picker already treats it as the default selection. Unconfigured picker keeps current behavior.

Verified live against a running dev instance: stale override → 403 refused; PUT excluded model → 400; explicit per-turn excluded model → 403; org default turn → 202 queued. Plus unit/integration tests (`turn-options`, `admin-resources`).

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yc-software/qm",
-  "version": "0.1.0",
+  "version": "0.1.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@yc-software/qm",
-      "version": "0.1.0",
+      "version": "0.1.3",
       "license": "MIT",
       "bin": {
         "qm": "dist/bin/qm.js"
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yc-software/qm",
-  "version": "0.1.2",
+  "version": "0.1.3",
   "license": "MIT",
   "description": "Control-plane CLI for portable QM deployments on Docker, Fly, and AWS.",
   "type": "module",
--- a/cli/templates/deployment/deployment.md
+++ b/cli/templates/deployment/deployment.md
@@ -158,21 +158,12 @@ A playground must be its **own deployment**, never a flag on a working org's
 instance: every visitor is an ordinary internal principal of the deployment's
 org, so anything granted or published at org scope — including org-granted
 credentials — is theirs. Grant nothing sensitive at org scope, connect no real
-connector credentials, and load no company data.
-
-Because a cleared cookie is a fresh identity, the per-principal brakes are
-soft; set the real ones in the same pass: `env.core.ORG_BUDGET_USD_PER_WINDOW`
-(the one hard spend ceiling), `BUDGET_USD_PER_WINDOW` and
-`RATE_LIMIT_PER_WINDOW` per principal, and after first boot pin a single model
-through the Admin page's base-model and web-UI model resources. The portal
-refuses to boot a playground alongside `PORTAL_COOKIE_DOMAIN`,
-`PORTAL_APPS_DOMAIN`, or `PORTAL_DEPLOYMENTS_ENABLED`, and refuses mint limits
-outside the claim store's bounds. Anonymous visitors are denied `/admin` and
-the connect and secret-drop flows. Nothing garbage-collects an abandoned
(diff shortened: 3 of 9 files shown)


16francej:
Fresh-context adversarial review of the picker enforcement ran; disposition:

**Fixed**
- *Org-default exemption was unreachable for the real web UI* (it always sends an explicit `model:` — the default selection is the org base model — so a picker excluding the org default refused every default turn): the enabled set is now the configured picker ∪ org default at all three entries (per-turn option, stored-override resolution, `PUT /v1/runtime-config`).
- *`keep: true` legacy migration could mint an unchecked selection*: it now skips migrating a legacy base model the picker excludes.
- *Test gap*: added an app-level test driving `app.turn` end-to-end (stale override → refused; explicit org default and inherit → queued) — the unit test alone survived swapping the wiring's arguments.

**Intended / documented, not changed**
- Fail-closed on picker change: pre-existing scope overrides outside a newly configured picker refuse web turns until re-picked or inherited — that's the point of the fix, but note it when configuring a picker on a live deployment. The picker is one flat org-wide list across harnesses; include ids for every approved harness you want usable on web.
- Cron/automation turns don't consult the web-UI picker (it's a web knob; Slack is likewise exempt). Post-fix, users can no longer *create* excluded overrides, so this only affects overrides that predate the picker.
- With no picker configured, `PUT /v1/runtime-config` still accepts any approved-harness-supported model (pre-existing; configuring the picker is the boundary the feature promises).

Re-verified live against a running dev instance after the changes: stale override → 403, PUT excluded → 400, explicit excluded → 403, explicit org default → 202, override-to-org-default → 200, inherit → 202.

PR #51, opened 2026-07-31
fix(slack): acknowledge terminal delivery after verified post

### LINKED ISSUE #44: Post the Slack reply before acking its recovery delivery
## Problem

`pollRun` acks a run's recovery delivery as soon as the run reaches a terminal state (`src/slack/core-bridge.ts:154-156`), which tombstones the durable copy. The reply itself is posted afterwards by the turn handler (`src/slack/turn-handler.ts:543,548`) through `postReply`, which is a plain `chat.postMessage` with retries disabled (`NO_RETRY`, `src/slack/config.ts:3`, applied at `src/slack/index.ts:47`) and carries no idempotency metadata.

If that single post attempt fails — a `chat.postMessage` rate limit, a transient 5xx, or a redeploy during the attachment upload window — the exception is swallowed, the run is already terminal, and the delivery is already stamped delivered. The user never receives the answer and nothing retries it.

The quarantine refusal path deliberately does the opposite, posting before it acks (`postThenAckRunDelivery`, `src/slack/refusals.ts:15-27`).

The current ordering is load-bearing for duplicate suppression: the live post carries no `qm_delivery` metadata, so the delivery poller's verify-first scan could not recognize it and would repost. Any fix needs to preserve that.

## Expected behavior

A run's reply is durably recoverable until it has actually been posted. A post failure leaves the delivery pending so the existing poller retries it, without introducing a duplicate for the common case where the live post succeeded.

## Acceptance criteria

- The recovery delivery is acked only after the reply has been posted successfully.
- The live post carries the same identifying metadata the delivery poller verifies against, so a retry cannot double-post.
- A failed live post leaves the delivery claimable, and the poller delivers it.
- Tests cover a failing `chat.postMessage` and assert the user still receives exactly one reply.

morluto:
Fixes #44

The Slack bridge was acknowledging a terminal run before the live reply was durably visible. A normal reply used a plain `chat.postMessage`, so an uncertain Slack failure could leave the recovery copy acknowledged with no reliable way to find an already accepted post.

This changes the terminal protocol so the live handler posts with `postWithVerify(..., "run:<id>")`, completes the text/update, attachment uploads, and task-list finalization, and only then starts the recovery-copy ack. Failed posts release the local claim and leave the durable delivery pending. The post-before-ack helper now lives with the delivery code and is shared by normal terminal replies and refusal delivery. Existing `ackByKey` tombstones are unchanged.

Attachment completion is also recorded per delivery index, including the enqueue race, so a recovery attempt does not upload files that the live attempt already completed. Requester-only channel refusals remain ephemeral by design; those refusal results do not create a durable recovery delivery.

Validation:
- 125 focused Slack, delivery, attachment, and run-result tests passed
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`

### DIFF
--- a/src/api/app-messaging.ts
+++ b/src/api/app-messaging.ts
@@ -49,6 +49,8 @@ export function createMessagingMethods(
   | "fulfillContextRequest"
   | "ackDelivery"
   | "ackDeliveryByKey"
+  | "recordDeliveryAttachment"
+  | "recordRunDeliveryAttachment"
   | "setRunDeliveryState"
   | "upsertDirectory"
   | "upsertChannels"
@@ -295,6 +297,12 @@ export function createMessagingMethods(
     },
     async ackDeliveryByKey(idempotencyKey) {
       await deps.deliveries.ackByKey(idempotencyKey, Date.now());
+    },
+    async recordDeliveryAttachment(id, index) {
+      await deps.deliveries.recordAttachment(id, index);
+    },
+    async recordRunDeliveryAttachment(runId, index) {
+      await deps.deliveries.recordAttachmentByKey(`run:${runId}`, index);
     },
     async setRunDeliveryState(runId, state) {
       const found = await deps.runs.setDeliveryState(runId, null, state);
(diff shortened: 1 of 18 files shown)


ReganBell:

→ #215

PR #62, opened 2026-07-31
The durable map accepts strings Postgres can't store

Saidheerajgollu:
Adding an ADR in `adrs/`, per CONTRIBUTING.

Short version: `createMemoryMap` and `createPostgresMap` disagree about what's storable.
The memory map takes any JS string; the Postgres map writes `JSON.stringify(value)` into
`jsonb`, and Postgres rejects a NUL and an unpaired surrogate. Every store behind
`artifactMap(...)` sits on that seam.

What makes it more than an edge case is that qm produces the bad string itself: `cleanName`,
`normalizeTitle` and the scheduler's `truncate` all clamp with a bare `.slice()`, which
strands half a surrogate pair when the cut lands mid-emoji. A 200+ character project name
with an emoji across character 200 is enough:

```
memory map (every test):     OK -> stored name ends with "bcd\ud83d"
postgres map (production):   THREW -> invalid input syntax for type json
```

I ran it against Postgres 18 using the repo's own `createPostgresMapFactory`, project store
and cron scheduler; `test/postgres-map.test.ts` passes clean against the same instance, so
this isn't the environment. I also chased the cron path to the end to check how bad it gets,
and it's milder than it first looks — `runTrigger`'s idempotency key absorbs the retry, so
the damage is one lost fire-log entry rather than a repeated turn. That's in the file too.

`headSlice`/`tailSlice` and `jsonbSafeStringify` already exist for exactly this, and four
places in the repo already treat both characters as unstorable — just not the layer they all
write through. Details, the cases I ran, and what I'd suggest are in the file.

Not a security report: nothing crosses a scope and the failures are loud and recoverable.
The reason I thought it was worth writing up is that it can only fail on a real deployment,
so the suite can't see it.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/62"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788127725&installation_model_id=19911&pr_number=62&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F62&signature=a56d305e5910c537f403b4951b58128dd8a245a804080a592b710adab8e3fde0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/adrs/durable-map-unstorable-strings.md
+++ b/adrs/durable-map-unstorable-strings.md
@@ -1,0 +1,67 @@
+# The durable map accepts strings Postgres can't store
+
+`createMemoryMap` and `createPostgresMap` don't agree on what a valid value is. The memory
+map takes any JS string; the Postgres map writes `JSON.stringify(value)` into a `jsonb`
+column, and Postgres rejects two things that are perfectly legal in a JS string: a NUL, and
+an unpaired surrogate.
+
+I ran this against a real Postgres 18 with the repo's own `createPostgresMapFactory`:
+
+```
+memory map:    put title with NUL          -> OK
+postgres map:  put title with NUL          -> unsupported Unicode escape sequence
+postgres map:  merge title with NUL        -> unsupported Unicode escape sequence
+postgres map:  put lone surrogate          -> invalid input syntax for type json
+                                              (22P02, "Unicode low surrogate must follow a high surrogate")
+```
+
+So every artifact store behind `artifactMap(...)` — crons, projects, skills, monitors,
+souls, keychain asks, the rest — can be handed a value that works in every test and throws
+in production.
+
+The part that makes this more than a theoretical edge case is that qm manufactures those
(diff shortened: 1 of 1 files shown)


ReganBell:

PR #66, opened 2026-07-31
Support TinyFish through OAuth MCP

KateZhang98:
Enable TinyFish as a per-person OAuth MCP connection in QM, without a skill or API key.

Requested QM behavior:
- identify the Slack speaker or scheduled-instruction owner
- use that person’s TinyFish OAuth connection
- inject search, fetch_content, run_web_automation, run_web_automation_async, and get_run into the shared harness tool set
- pass calls and results through QM approvals, external-content screening, and audit
- never share one participant’s connection with another

Live TinyFish verification completed: OAuth, tool discovery, search, fetch, synchronous automation, and asynchronous automation through get_run all succeeded.

The requested QM checks cover the full Slack-to-tool path for search, URL fetch, interactive automation, and explicitly backgrounded automation.

### DIFF
--- a/adrs/tinyfish-mcp.md
+++ b/adrs/tinyfish-mcp.md
@@ -1,0 +1,16 @@
+# TinyFish MCP
+
+Could QM support an org-configured remote MCP server that people connect to with OAuth?
+
+We want to connect `https://agent.tinyfish.ai/mcp` without a skill or API key. For now we need these TinyFish tools:
+
+- `search` for external knowledge and finding pages
+- `fetch_content` for reading supplied URLs
+- `run_web_automation` for interactive website tasks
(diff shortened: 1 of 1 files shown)


KateZhang98:
@16francej could you review this plan before implementation, especially the per-speaker OAuth ownership and injecting the TinyFish tools through QM’s shared harness tool path? The TinyFish OAuth MCP endpoint and requested tools were verified live.

ReganBell:
Thanks for the detailed write-up — the end-to-end OAuth test against the endpoint was genuinely useful signal. We are consolidating all vendor-specific MCP asks under #75: the plan is a generic MCP connector slot rather than first-party integrations per vendor. TinyFish would ride that as a recipe. Closing in favor of #75 — please weigh in there.

PR #69, opened 2026-08-01
Add a web tool for search and page reading

nickscamara:
Today the agent reads the web by shelling out to curl, and the browse skill points it there. That works, but it returns raw HTML, misses anything rendered client side, and skips the security screening every other external input goes through. This adds a purpose built `web` tool.

**What's new**

* `web` tool with two actions: `search` (recency, domain and news filters) and `scrape` (one url, returns markdown)
* Works without an API key, so a fresh install has web access on its first turn
* `FIRECRAWL_API_KEY` raises the rate limits. `FIRECRAWL_BASE_URL` targets a self hosted instance or an egress proxy
* Results route through the external content screener, so a scraped page is checked for prompt injection before the model sees it. `execute` output is not screened
* A new `harnessToolOptions` helper replaces the tool option fields that all four harnesses previously hand copied
* The browse skill now points at `web` for reading pages and reserves the browser for interaction

**No vendor lock in**

* The tool is called `web`
* Providers implement a two method `WebToolDeps` interface, wired at a single call site in `wiring.ts`
* Swapping in another vendor, an internal search index, or a self hosted crawler means writing `search` and `scrape` and changing that one line. The tool, all four harnesses, the prompt, and the security path stay untouched
* Firecrawl is the default because it is the one that works with zero configuration, not because anything is bound to it

**Testing**

* 29 tests covering request shaping, response mapping, keyless headers, error paths and screening
* Verified the client against the live API, including keyless search and scrape

**Caveats**

* A `web` call is core side, so it bypasses the org's sandbox egress allowlist and audit sink. Happy to wire that up before merge
* Keyless is metered per IP per day rather than per account, so a shared deployment shares one allowance

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,6 +6,9 @@ HARNESS_SECURITY_POSTURE=auto
 #OPENAI_API_KEY=sk-...
 #OPENROUTER_API_KEY=sk-or-...
 
+#FIRECRAWL_API_KEY=fc-...
+#FIRECRAWL_BASE_URL=https://api.firecrawl.dev/v2
+
 ORG_ID=acme
 PORT=8080
 
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yc-software/qm",
-  "version": "0.1.4",
+  "version": "0.1.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@yc-software/qm",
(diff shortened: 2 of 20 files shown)


time-attack:
Thanks for this, @nickscamara — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/839 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #70, opened 2026-08-01
Let QM use Migma.ai for email campaigns

AdamSEY:
## What

Proposes using Migma.ai as QM's email campaign connection.

Migma lets agents create, edit, validate, send, schedule, and track branded
emails. Through Migma's OAuth MCP server, a QM user could run that workflow from
Slack or web: prepare a campaign, review it, approve the send, then read
delivery and click results back in the same conversation.

## Why

I noticed the TinyFish OAuth MCP proposal. Migma can use similar per-person
connection ownership, but serves a different job: email creation, sending,
audience work, and campaign tracking. Sends and audience changes remain behind
QM approval and audit. Migma remains the email and tracking source of truth, so
QM needs no duplicate campaign database or UI.

## Checks

- `git diff --cached --check`
- Confirmed one-file ADR-only diff

### DIFF
--- a/adrs/migma-oauth-mcp.md
+++ b/adrs/migma-oauth-mcp.md
@@ -1,13 +1,23 @@
-# Migma over OAuth MCP
+# Use Migma.ai for email campaigns in QM
 
-I saw the TinyFish OAuth MCP proposal. Migma fits the same connection seam, but
-the job is different: TinyFish brings web tools; Migma handles email work and
-includes consequential writes.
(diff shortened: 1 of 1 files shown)


ReganBell:
Thanks for the proposal. We are consolidating vendor-specific MCP asks under #75 — the direction is a generic MCP connector slot, and Migma would connect through that rather than as a first-party integration. Closing in favor of #75.

PR #73, opened 2026-08-01
A self-hosted company brain over MCP

time-attack:
garry tan texted me saying sina we have to add gbrain implementation to this so i whipped up a nice version rq. This is it yall.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/73"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788152074&installation_model_id=19911&pr_number=73&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F73&signature=1519bb237eeaeae690d2796ccf6f8934dd4a48c3ea1c5f270467acdac9647ccb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #78, opened 2026-08-01
Add an optional gbrain memory backend

time-attack:
Follow-up to #73 with the actual implementation.

## What this does

Lets a deployment point QM at a gbrain instance, so `query()` can fall back to the company brain when a scope's own notebook doesn't have the answer. The notebook is still the only writer of scope memory; this only reads.

The notebook is the right shape for "what do I know about this person I'm talking to" and the wrong shape for "what does anyone here know about this customer" — it's a few hundred bullets per scope with substring matching. That second question is what a brain answers.

## Shape

A decorator around `MemoryService`, installed at the one wiring line. `recall()` is untouched, so the per-turn path never goes to the network. Only `query()` changes: it asks the brain when the local notebook didn't already fill the limit, and merges behind the local hits, deduped.

Off unless all four of `GBRAIN_MCP_URL`, `GBRAIN_ISSUER_URL`, `GBRAIN_CLIENT_ID` and `GBRAIN_CLIENT_SECRET` are set — with none set, `createGbrainMemory` returns the base service by identity. Setting some but not all is a boot error rather than a half-configured brain, and the URLs must be https off localhost.

Plain `fetch` against gbrain's OAuth 2.1 client-credentials endpoint and its HTTP MCP `tools/call` surface, so no new dependency. Token is cached and single-flight, one deadline covers mint + request + 401 retry, SSE frames are reassembled and matched to the request id, bodies are size-capped, redirects refused. The token is requested with **read** scope only, so a leaked deployment credential can't write to the brain.

It never throws at the memory caller: a brain that is down, slow, or returning garbage degrades to notebook-only behaviour and reports through the error log.

## Isolation

Results are filtered before they reach a turn. A row is dropped unless its slug is outside the `qm/` namespace (shared org content — the reason to consult a brain) or inside the asking scope's own prefix. Rows without a slug are dropped because visibility can't be established, and the slug is normalised (leading slashes, doubled separators, case, whitespace) before the check so a server that canonicalises differently than it stored can't pass another scope's page off as shared. Deleting the filter fails the isolation tests.

## What I deliberately left out

The first version also mirrored each scope's notebook into the brain, so captured facts became searchable org-wide. Review found the off-box copy brought problems the copy itself created: deletions propagated best-effort only, a failed mirror was never retried, and the per-scope ordering queue lived in process memory — which this codebase explicitly doesn't treat as durable, since core runs multi-instance. The failure mode is a fact a scope deleted locally persisting, and resurfacing, in the brain.

None of that is inherent to *consulting* a brain, so I cut it. Write-back is worth doing, but it needs durable delivery (pg-boss, retries, drain on shutdown) and an explicit deletion contract — its own change, not a footnote to this one.

## Review

Two independent clean-room reviews (Claude Fable 5 and codex gpt-5.6-sol) against an instruction-stripped copy of the tree, neither told who wrote it. Round one found a cross-scope leak (`search()` ignored `scopeId` entirely), that only `capture()` mirrored so deletions persisted, a mirror ordering race, and that the parser read `excerpt`/`content` when gbrain actually returns `chunk_text` — so every remote hit had been degrading to a bare title. All fixed or, for the mirror, cut.

## Testing

`npm test` 3730 tests / 0 failures. `npm run typecheck`, `lint:ox`, `format:check` clean. 18 new tests cover scope isolation, the merge/limit/dedupe logic, token caching and 401 re-mint, SSE and JSON framing, oversized bodies, and fail-soft behaviour.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/78"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788156754&installation_model_id=19911&pr_number=78&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F78&signature=de65dcbba5a36325aac37ba122f32ee961a6c843fc85106176ab9e1be9669b5b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


time-attack:
Context for reviewers: this replaces #73 (closed), which described the same integration as a note. The implementation here reads from the brain only — the write-back mirror that note mentioned was cut after review, since propagating deletions correctly needs durable delivery and an explicit deletion contract rather than best-effort mirroring from process memory.

PR #80, opened 2026-08-01
Two small gaps in the command redactor

Dhevenddra:
Adding an ADR for this rather than a diff, per CONTRIBUTING.

Was reading through `redactCommand` in `src/sandbox/exec-process-session.ts`, the function that scrubs a command before it lands in logs or audit, and hit two small holes.

First, base64url values don't get masked. `createSecretValueMasker` makes redaction needles for the raw value, the percent-encoded value, and standard base64, but not base64url (`-_`, usually unpadded). So a secret that appears in a command in URL-safe encoded form, like a JWT-shaped token, gets logged in the clear even though the plain base64 form would have been caught.

Second, the `--with-token` line does nothing. `.replace(/(--with-token\b)/gi, "$1")` replaces the match with itself, so it's a no-op. Either it was meant to redact the token after the flag, or it's leftover and can go.

Neither is a live exploit. It's defense in depth for the audit trail, not a boundary. But both are cheap to close, and the audit trail is worth keeping clean. Happy to send the change if you'd like it.

### DIFF
--- a/adrs/redact-command-gaps.md
+++ b/adrs/redact-command-gaps.md
@@ -1,0 +1,20 @@
+# A couple of gaps in the command redactor
+
+Was reading through `redactCommand` in `src/sandbox/exec-process-session.ts`, the function
+that scrubs a command before it gets logged or audited, and noticed two small holes that make
+it do less than it looks like it does.
+
+First, base64url values don't get masked. `createSecretValueMasker` builds redaction needles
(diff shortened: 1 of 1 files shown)


ReganBell:

Dhevenddra:
Thanks for the co-author credit on both of these. Happy either way as long as the splitter stops cutting emoji in half. I'll keep poking at qm.

PR #94, opened 2026-08-01
Resolve shared file handles that live only in the artifact store

Restry:
A file uploaded through the web UI's Files page never reaches the agent. Upload creates an ACL grant, that grant materializes as a `shared/<name>` handle, and the handle gets listed in the system prompt under "Files shared with you" — but `read shared/<name>` returns null, which the model sees as `[no such file]`. So the agent is told the file is there and then can't open it.

The cause is that two things create file grants and they use different stores. The `write(path, data, share)` tool writes bytes to the workspace *and* the artifact store under the real workspace path, so its handles resolve. `uploadFileForViewer` and inbound message attachments write only to the artifact store, under the logical path `artifacts/<id>/<name>`, and the shared-handle read path consulted the workspace store alone. Message attachments don't show the bug because they skip handles entirely — they're materialized into `inbox/` at turn time.

I noticed this while reading the codebase to answer a question about whether QM could carry per-project context for someone with a lot of repos, and it looked like a wiring gap rather than a deliberate boundary — the grant, the handle, and the system-prompt line are each covered by their own test, so four of the five steps are clearly intended.

I'm aware `CONTRIBUTING.md` asks for a written note in `adrs/` rather than code. The description above is the part worth reading; the patch is attached in case it saves you the work, and I'm happy for it to be closed and redone your way.

---

### What the patch does

`read` falls back to the artifact store for handles in the artifact namespace, keeping the existing text-inline / binary-materialize split.

The fallback is namespace-gated, so a workspace-backed share can never be served from the artifact store's snapshot of it. That matters because `put` returns the existing row for a deterministic id, so a second `write` to one path updates the workspace and leaves the artifact holding the first version forever. Gating on the namespace gives the two stores non-overlapping responsibility instead of a precedence rule with a staleness caveat.

The resolver also re-asserts that the row it got back is the `(ownerScopeId, path)` it asked for, rather than trusting result ordering — the schema's index on `(owner_scope_id, path)` isn't unique and neither implementation orders results.

Authorization is unchanged and stays upstream: the resolver only ever queries the exact tuple `handlesForAudience` already blessed.

Bytes are capped at 10MB. `files.open` reports `sizeBytes` before the read, but on the S3 backend that's `ContentLength ?? 0`, so the cap is enforced by the bounded collector too rather than trusting the advertised size — uploads can be up to 1GB, and a missing `ContentLength` would otherwise fail open on the largest possible object.

`artifactPath`/`isArtifactPath` name the namespace both producers were spelling out by hand.

### Unrelated thing I ran into

Three source files carried raw NUL bytes where a `\0` escape was meant — `src/files/file-artifact-store.ts`, `src/surface-cache/surface-cache.ts`, `test/live-slack/gallery.ts`. They're separator characters in composite keys and a regex character class, so behaviour is identical, but the raw bytes make the files read as binary to `file(1)`, to editors, and to review tooling, and they hide the separator: `` `${a}<NUL>${b}` `` displays as `` `${a}${b}` ``. That misled a reviewer of this change into reporting the artifact store's lookup key as an unseparated concatenation that could alias across scopes. It was never that — but it cost real time, so I converted them to `\0`.

Happy to split this into its own PR if you'd rather.

### Something I deliberately left alone

`uploadFileForViewer` does no security screening, while the message-attachment path screens and drops strict-verdict content *before* `registerArtifact`, so blocked bytes never become a grantable artifact. This change turns those unscreened upload bytes from human-download-only into model-readable.

I didn't bundle a fix because it's a different subsystem and needs a product call — what does the HTTP upload return when content is blocked? Flagging it rather than quietly widening the channel.

### Verification

`npm run typecheck` and `npm run lint` clean. Full root suite: 3717 tests, 0 failures. The three new behavioural tests fail without the `primitives.ts` change.

No screenshot — this is core plumbing with no rendered surface.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/94"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788190266&installation_model_id=19911&pr_number=94&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F94&signature=6165efeaeadea55a592a218785c508c51697d7f0e70a78d10b3c369074468b30"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Restry:
Heads-up for review: `src/files/file-artifact-store.ts` and `test/live-slack/gallery.ts` show as `+0/-0` with no rendered diff. That's not an empty change — GitHub classifies the *pre-image* as binary because it contains raw NUL bytes, which is exactly what those two hunks fix. The actual edits are:

- `file-artifact-store.ts` — two raw NULs to `\0` in the `resolveByOwnerPaths` key, plus the new `artifactPath`/`isArtifactPath` helpers
- `gallery.ts` — three raw NULs to `\0` inside regex character classes

`git show 2aa43b0 -- src/files/file-artifact-store.ts` renders fine locally once the file is text on both sides. Every file in the repo is clean of raw NULs after this; the only remaining matches are the two PNGs, which are legitimately binary.

ReganBell:

PR #110, opened 2026-08-02
Let a deployment point model providers at its own gateway

Restry:
The in-process Pi harness can only ever reach the vendor endpoints. `pi-ai` bakes `baseUrl` into each provider's model record and nothing read an override, so `resolveModel` always returns `https://api.anthropic.com` / `https://api.openai.com/v1` / `https://openrouter.ai/api/v1`.

The child-process harnesses are already in a different position. `config.ts` forwards `ANTHROPIC_BASE_URL` into `claudeProcessEnv` and `OPENAI_BASE_URL` into `codexProcessEnv`, and each vendor SDK honours them. So on a deployment behind a gateway, Claude Code and Codex work and Pi doesn't — a difference in reach that the harness abstraction is otherwise good at hiding. I hit this running QM against a self-hosted Anthropic-compatible gateway: switching the org to Pi silently pointed every request back at the vendor.

Config now parses `ANTHROPIC_BASE_URL`, `OPENAI_BASE_URL` and `OPENROUTER_BASE_URL` — the same variables the SDKs read — and `buildApp` hands them to the model registry.

### Where the override lands

In `builtinModel`, not in `resolveModel`'s return. `MODEL_REGISTRY` clones some entries from a template (`claude-opus-5` is a clone of `claude-opus-4-8`), and `cloneModel` spreads the template — so overriding the template covers the clone, while overriding only the direct return path would leave clones pointing at the vendor while their template moved. Everything that issues a request resolves through `builtinModel`, so the agent loop, judge, compaction, title, ack and detect calls all follow one setting.

The value replaces the vendor base URL verbatim. Providers disagree about whether the version prefix belongs in it — anthropic omits `/v1` and pi-harness appends `/v1/messages`, openai includes it — and matching each provider's own convention is what keeps these variables interchangeable with the SDKs that already read them.

### Injection rather than a direct env read

`pi-models.ts` can't import `config.ts`: config already imports pi-models, so it would be a cycle. And reading `process.env` from pi-models is what the repo's own lint rule exists to prevent. So config parses, `buildApp` injects via `setProviderBaseUrls`, and pi-models keeps the resolved value. Threading it through `resolveModel`'s ~15 call sites across 8 files was the alternative; it seemed a lot of surface for a value that is fixed for the process lifetime.

### Admin key validation

`putModelProvider` posted to a hardcoded vendor URL to check a key. Against a gateway that never sees the vendor's key, that fails a key which actually works, so the admin UI refuses to save it. It now resolves through the same setting.

### Verification

`npm run typecheck` and `npm run lint` clean; full root suite 3715 tests, 0 failures. New tests cover the parse (including trailing-slash and whitespace normalization) and the resolution, with the clone case pinned explicitly — that one fails if the override moves to `resolveModel`.

Beyond the suite, this is running on a live single-user deployment pointed at an Anthropic-compatible gateway. With `HARNESS=pi` the agent completes turns and drives the sandbox through the gateway; both Pi and Claude Code are approved on the same org and both execute against the same gateway with no vendor key present anywhere. Before the change the same instance could only run Claude Code.

No screenshot — this is config plumbing with no rendered surface. The only UI it touches is the admin model-provider form, which behaves as it did except that a gateway-issued key now validates.

### Not addressed

`secret-schema.ts` still requires `ANTHROPIC_API_KEY` when the base model is Anthropic. A gateway deployment typically carries its token in `ANTHROPIC_AUTH_TOKEN` (which `claudeChildEnv` already forwards), so it ends up setting both to the same value. Accepting either felt like a separate call about what the schema is asserting, so I left it.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/110"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788227051&installation_model_id=19911&pr_number=110&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F110&signature=311b4e79e1dcfcbfc046f6cb99027271e77b44d3e85d5849b894f44ad4c8cc31"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/api/routes/admin/model-providers.ts
+++ b/src/api/routes/admin/model-providers.ts
@@ -1,24 +1,34 @@
-import { isModelProvider, type ModelProvider } from "../../../model/pi-models.ts";
+import { isModelProvider, providerBaseUrl, type ModelProvider } from "../../../model/pi-models.ts";
 import { selectableModelCatalog } from "../../../model/model-catalog.ts";
 import { sendJson } from "../../http.ts";
 import type { ApiCtx } from "../route.ts";
 import { audit, authorizeAdmin, orgScope } from "../shared.ts";
 
-const VALIDATION_REQUESTS: Record<ModelProvider, { url: string; headers: (apiKey: string) => Record<string, string> }> =
-  {
-    anthropic: {
-      url: "https://api.anthropic.com/v1/models",
-      headers: (apiKey) => ({ "x-api-key": apiKey, "anthropic-version": "2023-06-01" }),
-    },
-    openai: {
-      url: "https://api.openai.com/v1/models",
-      headers: (apiKey) => ({ authorization: `Bearer ${apiKey}` }),
-    },
-    openrouter: {
-      url: "https://openrouter.ai/api/v1/key",
-      headers: (apiKey) => ({ authorization: `Bearer ${apiKey}` }),
-    },
-  };
+const VALIDATION_REQUESTS: Record<
+  ModelProvider,
+  { baseUrl: string; path: string; headers: (apiKey: string) => Record<string, string> }
+> = {
+  anthropic: {
+    baseUrl: "https://api.anthropic.com",
+    path: "/v1/models",
+    headers: (apiKey) => ({ "x-api-key": apiKey, "anthropic-version": "2023-06-01" }),
+  },
+  openai: {
+    baseUrl: "https://api.openai.com/v1",
+    path: "/models",
+    headers: (apiKey) => ({ authorization: `Bearer ${apiKey}` }),
+  },
+  openrouter: {
+    baseUrl: "https://openrouter.ai/api/v1",
+    path: "/key",
+    headers: (apiKey) => ({ authorization: `Bearer ${apiKey}` }),
+  },
+};
+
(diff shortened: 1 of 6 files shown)


time-attack:
Thanks for this, @Restry — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/840 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #132, opened 2026-08-02
Make approvals single-use and bound to one action

Arthur9293:
## Summary

Propose making approvals single-use and bound to one specific action inside a limited scope.

## Scope

ADR text only, following the repository contribution guide. No code changes.


ReganBell:

→ #219

PR #140, opened 2026-08-02
adr: avoid submitting while IME composition is active

foxytanuki:
I hit this while using QM with Japanese input: Enter used to confirm an IME conversion can submit partial text and leave the rest in the composer.

The short proposal is in `adrs/web-composer-ime.md`. I originally opened this as an implementation PR before noticing the contribution guidance, so I changed this PR to the informal proposal instead.

The tested implementation is preserved on [`foxytanuki:hotfix/web-composer-ime`](https://github.com/foxytanuki/qm/tree/hotfix/web-composer-ime) if it's useful later.


ReganBell:

PR #147, opened 2026-08-03
Recover PostgreSQL session pools after connection failures

### LINKED ISSUE #136: core: Postgres pool death ('Connection terminated unexpectedly') never recovers without a machine restart (fly target, Managed Postgres basic)
**Setup:** qm 0.1.4, target=fly, region iad, Fly Managed Postgres (basic plan), HARNESS=pi, SANDBOX_BACKEND=sprites (default).

**Symptom:** Roughly every 1–2 hours, core loses its Postgres connections and never recovers. Logs fill with:

```
[cron-queue] pg-boss error: Connection terminated unexpectedly
[cron-queue] pg-boss error: [object Object]
[server] 500 GET /v1/contexts?...: Error: Connection terminated unexpectedly
[server] 500 GET /v1/sessions?...: Error: Connection terminated unexpectedly
[monitor] tick failed: Error: Connection terminated unexpectedly
[scheduler] tick failed: Error: Connection terminated unexpectedly
```

Every API request 500s from then on (web UI shows "internal server error" on chat send). The MPG cluster itself reports `ready` the whole time — only core's pool is wedged.

**Workaround:** `fly apps restart <prefix>-core` fixes it instantly, until the next occurrence (3 occurrences in ~4 hours today).

**Expected:** the pg pool (and pg-boss) should reconnect with backoff after connection loss; the `/healthz` check should probably also exercise a DB round-trip so Fly can restart the machine automatically when this happens.

Happy to provide more logs or run a debug build.

IRONICBo:
## Summary

- retire PostgreSQL pools after connection-level failures so later requests can recover without restarting QM
- route session, directory, and memory transactions through the recoverable pool boundary
- make session-backed health checks bounded and close PostgreSQL session pools during shutdown
- preserve the original transaction error when rollback also fails

## Validation

- `npm run typecheck`
- `npm run lint -- --quiet`
- `npm run lint:ox`
- focused route and pool tests: 37 passed, 1 skipped without `DATABASE_URL`
- `npm run test:pg`: 148 passed against PostgreSQL 16
- live PostgreSQL restart: health failed while unavailable, recovered after restart, and retained the existing session
- `npm test`: 3,581 passed, 132 skipped; 10 sandbox migration failures reproduce unchanged on `upstream/main` in the same environment

Fixes #136

### DIFF
--- a/src/api/routes/index.ts
+++ b/src/api/routes/index.ts
@@ -25,8 +25,30 @@ import { deploymentLayerRoutes } from "./deployment-layer.ts";
 import { egressAuditRoutes } from "./egress-audit.ts";
 import { authBrokerRoutes } from "./auth-broker.ts";
 
+const HEALTH_TIMEOUT_MS = 1_500;
+
 export const rawRoutes: ReadonlyArray<Route<BaseCtx>> = [
-  { method: "GET", path: "/healthz", auth: "public", handle: ({ res }) => sendJson(res, 200, { ok: true }) },
+  {
+    method: "GET",
+    path: "/healthz",
+    auth: "public",
+    handle: async ({ res, deps }) => {
+      let timer: NodeJS.Timeout | undefined;
+      try {
+        await Promise.race([
+          deps.sessions?.health?.(),
+          new Promise<never>((_, reject) => {
+            timer = setTimeout(() => reject(new Error("health check timed out")), HEALTH_TIMEOUT_MS);
+          }),
+        ]);
+        sendJson(res, 200, { ok: true });
+      } catch {
+        sendJson(res, 503, { ok: false });
+      } finally {
+        if (timer) clearTimeout(timer);
+      }
+    },
+  },
   {
     match: (m, p) => (m === "GET" || m === "POST") && p.startsWith(GIT_HTTP_BROKER_PREFIX),
     auth: { aud: "credential-broker" },
--- a/src/directory/postgres-directory-store.ts
+++ b/src/directory/postgres-directory-store.ts
@@ -1,6 +1,6 @@
 import { orgId as configOrgId } from "../config.ts";
 import { createHash } from "node:crypto";
-import { createPgPool, type PoolClient, withPgTransaction } from "../persistence/pg-pool.ts";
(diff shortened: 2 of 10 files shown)


rajpratham1 (approved):
This PR significantly improves the resilience of the PostgreSQL layer by introducing automatic pool recovery, transaction abstraction, health checks, graceful shutdown behavior, and comprehensive regression tests. The changes are cohesive and address a real operational issue where transient PostgreSQL failures could permanently poison the application's connection pool.

What looks good
Automatic pool recovery
Introduces retirement of failed pools instead of continuing to use broken connections.
Detects connection-level failures and transparently recreates pools.
Handles idle client errors emitted by pg.Pool.
Centralized transaction API
Replaces repeated manual transaction handling with a shared transaction() helper.
Reduces duplicated BEGIN/COMMIT/ROLLBACK logic across multiple stores.
Makes transaction behavior consistent everywhere.
Graceful shutdown
Prevents reopening pools after shutdown.
Properly drains retired pools before closing them.
Ensures close() behaves deterministically.
Health endpoint improvements
/healthz now validates session-store connectivity rather than always returning success.
Includes timeout protection.
Correctly avoids leaking database details in error responses.
Session store improvements
Adds optional health() and close() interfaces.
Runtime shutdown now attempts to close the session store without blocking indefinitely.
Excellent test coverage
Recovery after idle client failures.
Connection-level failures.
Transaction failures.
Pool retirement.
Shutdown behavior.
Health endpoint success/failure/timeout.
Pool recovery edge cases.
Code quality

The refactoring appears consistent:

repeated withPgTransaction(await pool(), ...) calls are replaced with db.transaction(...)
common functionality is centralized instead of duplicated
connection recovery logic is isolated inside pg-pool.ts rather than spread throughout callers

That makes the design easier to maintain.

Minor observation (non-blocking)

The implementation currently detects recoverable failures using PostgreSQL error codes together with regex matching on error messages. While practical, string-based message matching can be brittle if drivers or upstream libraries change wording. If additional failure modes appear in the future, it may be worth centralizing the detection policy even further or relying primarily on structured error metadata where available. This is an enhancement suggestion rather than a blocker.

ReganBell:

→ #220

PR #149, opened 2026-08-03
fix: catch transient getRun failures in postgres waitFor poll

### LINKED ISSUE #57: Handle PostgreSQL poll failures in run waitFor
`createPostgresRunStore().waitFor()` starts a polling interval whose `getRun(runId)` rejection is not handled:

```ts
void getRun(runId).then((r) => {
  if (r && isTerminal(r.status)) finish(r);
});
```

If Postgres is temporarily unavailable while a caller is waiting, the query rejects outside the returned `waitFor` promise. On current Node versions that is an unhandled rejection and can terminate the process instead of letting `waitFor` time out or report the database error normally.

The interval also leaves its timeout scheduled after `finish`, although it is unref'd.

Could the poll catch transient query failures (or reject `waitFor` deliberately) and clear both timers on settlement?

SomSamantray:
## Issue

`waitFor`'s polling loop in `postgres-run-store.ts` called `getRun(runId).then(...)` with no `.catch()`. A transient query failure (e.g. a brief Postgres blip) became an unhandled rejection outside the returned promise, which can crash the process instead of letting `waitFor` time out normally. `finish()` also only cleared the poll interval, leaving the timeout timer pending.

## Fix

- Poll tick now catches a rejected `getRun` call, logs it, and lets the next tick (or the existing timeout) handle it — no behavior change on the success path.
- `finish()` clears both the poll interval and the timeout timer.
- Added a test that forces a rejection mid-poll and asserts no unhandled rejection occurs and `waitFor` still settles via its timeout.

Fixes #57

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/149"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788340491&installation_model_id=19911&pr_number=149&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F149&signature=4a9fb1fc6562ecf808f29ca3fc0a37149bc1086bdbf73c4fb3808c41f4856ef7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/runs/postgres-run-store.ts
+++ b/src/runs/postgres-run-store.ts
@@ -290,6 +290,7 @@ export function createPostgresRunStore(connectionString: string, opts?: { maxCla
           if (done) return;
           done = true;
           clearInterval(poll);
+          clearTimeout(timer);
           events.off(runId, onSettle);
           resolve(r);
         };
@@ -298,9 +299,13 @@ export function createPostgresRunStore(connectionString: string, opts?: { maxCla
         }
         events.once(runId, onSettle);
         const poll = setInterval(() => {
-          void getRun(runId).then((r) => {
-            if (r && isTerminal(r.status)) finish(r);
-          });
+          void getRun(runId)
+            .then((r) => {
+              if (r && isTerminal(r.status)) finish(r);
+            })
+            .catch((err: unknown) => {
+              console.error(`[postgres-run-store] waitFor poll for run ${runId} failed transiently:`, err);
+            });
         }, 250);
         poll.unref?.();
         const timer = setTimeout(() => {
--- a/test/postgres-store.test.ts
+++ b/test/postgres-store.test.ts
@@ -894,6 +894,31 @@ test("pg run store: enqueue dedup, atomic one-per-session claim, fencing, ledger
 });
 
 test(
+  "pg run store: waitFor survives a transient poll failure without an unhandled rejection",
+  { skip },
+  async () => {
+    const { runs, close } = createPostgresRunStore(URL!);
+    let unhandled: unknown;
+    const onUnhandledRejection = (err: unknown): void => {
+      unhandled = err;
+    };
+    process.once("unhandledRejection", onUnhandledRejection);
+    try {
+      const r = (await runs.enqueue({ sessionId: "sWaitFor", request: turn("x") })).run;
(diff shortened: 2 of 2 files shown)


rajpratham1 (approved):
This PR fixes a reliability issue in waitFor() by handling transient failures from the polling getRun() call instead of allowing them to surface as unhandled promise rejections. The change is small, targeted, and backed by an appropriate regression test.

What the change does
✅ Wraps the polling getRun(runId) promise chain with a .catch(...).
✅ Treats polling failures as transient instead of crashing the polling loop.
✅ Continues polling until either:
the run reaches a terminal state, or
the existing timeout expires.
✅ Logs the transient polling error to aid diagnosis.
✅ Adds a regression test that:
intentionally closes the PostgreSQL connection pool during polling,
verifies no unhandledRejection event is emitted,
confirms waitFor() still resolves through its normal timeout behavior.
Review

Strengths

✅ Fixes a real asynchronous error-handling gap.
✅ Preserves existing waitFor() semantics rather than changing observable behavior.
✅ The regression test accurately reproduces the failure scenario by forcing getRun() to reject mid-poll.
✅ Prevents noisy unhandled promise rejections that could otherwise terminate Node.js processes depending on runtime configuration.
Minor suggestion (non-blocking)

The current implementation logs every transient polling failure via console.error. If the database remains unavailable for an extended period, this could generate a large number of identical log entries (every 250 ms). Consider rate-limiting or deduplicating these messages, or using the project's structured logging facility if one exists. This would improve operational visibility without changing the recovery behavior.

ReganBell:

PR #155, opened 2026-08-03
Per-user model subscriptions via OAuth connectors (Anthropic + OpenAI)

roham:
## What this is

A feature request contributed as an ADR per CONTRIBUTING.md, with implementation pointers from reading the source.

## Summary

Both subscription-capable harnesses are org-centralized today:

- **Claude harness**: supports subscription billing via `CLAUDE_CODE_OAUTH_TOKEN` passthrough to the Claude Code subprocess — but the token is sourced from core env, so exactly one subscription identity exists per deployment.
- **Codex harness**: falls back to `OPENAI_API_KEY` because the Codex CLI cannot do browser OAuth in a container; ChatGPT subscriptions authenticate via OAuth, so per-user OpenAI subscriptions are unreachable the same way.

Organizations whose employees already pay for individual Claude Pro/Max and ChatGPT Plus/Pro seats get two bad options: share one person's subscription (rate limits included) org-wide, or buy API credits for usage employees already have.

## Ask

Make Anthropic and OpenAI first-class keychain connectors alongside the existing seven (Google, Slack, Notion, Linear, Dropbox, GitHub, X): user completes the provider's OAuth flow from the keychain page, the scoped token lands in the existing encrypted credential store, and each harness resolves its subscription credential per-scope with fallback to the core-env token / base-model key. Org defaults remain for scopes without a connected subscription.

The ADR notes why this looks architecturally consistent — per-scope credential storage, grant-scoped sandbox env injection, and user-initiated OAuth flows all exist today.

Happy to test against a live team deployment.

### DIFF
--- a/adrs/2026-08-03-per-user-claude-oauth-connector.txt
+++ b/adrs/2026-08-03-per-user-claude-oauth-connector.txt
@@ -1,0 +1,34 @@
+Feature request: per-user model subscriptions via OAuth connectors (Anthropic + OpenAI).
+
+Context: QM's harnesses each have a subscription-auth path that is today
+org-centralized:
+
+- The Claude harness passes CLAUDE_CODE_OAUTH_TOKEN through to the Claude Code
+  subprocess (src/harness/claude-harness.ts), enabling billing against a
+  Claude Pro/Max subscription. The token is sourced from the core process
+  environment (src/config.ts) — one subscription identity per deployment.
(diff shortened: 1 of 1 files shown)


iqbalbhatti49 (approved):

16francej:
Thanks for the ADR. This landed in #707 (per-user Claude/ChatGPT subscription auth, with an org-wide "Individual authorization for AI usage" toggle under admin > Governance), built on the subscription-auth branch from #690. Closing in its favor.

PR #164, opened 2026-08-03
Read local Postgres settings from the assembled env, not process.env

16francej:
## Problem

`dev-instance up` failed with `could not connect to DATABASE_URL: password authentication failed for user \"postgres\"` on a machine whose `~/.config/qm/dev.env` sets `DEV_INSTANCE_POSTGRES_PORT=55433`.

The CLI spawns the supervisor with only the caller's env (`scripts/dev/cli.ts`); dev.env is merged into `assembled.env` later by `assembleEnv()`. But `ensureLocalPostgres()` read `DEV_INSTANCE_POSTGRES_*` from `process.env`, so dev.env settings never reached it. It defaulted to port 55432 — which on this machine belongs to a *different* Postgres container with a different password — while the `qm-dev-postgres` container name check passed, producing the misleading auth failure.

## Fix

`ensureLocalPostgres()` takes the env record and the supervisor passes `assembled.env` (which already layers caller env over dev.env). No default-to-`process.env` fallback, so the resolution order is single-sourced.

## Verified

- Before (on `main`): `up` fails as above unless `DEV_INSTANCE_POSTGRES_PORT=55433` is exported manually.
- After: `up` boots clean with no manual export — canary round-trips, core runs `store=postgres` against the dev.env-configured container. Torn down after.
- `dev-cli-lib` and `dev-supervisor-child` tests, typecheck, lint, format all pass.

Found while live-QAing #160.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/164"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788382958&installation_model_id=19911&pr_number=164&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F164&signature=c8d05e49bdc713b1be037863dd6de860aae95b2b66230de453919ea89c5e62d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/scripts/dev/lib/postgres.ts
+++ b/scripts/dev/lib/postgres.ts
@@ -67,12 +67,16 @@ export interface LocalPostgres {
   url: string;
 }
 
-export async function ensureLocalPostgres(worktree: string, log: (msg: string) => void): Promise<LocalPostgres> {
-  const container = process.env.DEV_INSTANCE_POSTGRES_CONTAINER || "qm-dev-postgres";
-  const port = process.env.DEV_INSTANCE_POSTGRES_PORT || "55432";
-  const image = process.env.DEV_INSTANCE_POSTGRES_IMAGE || POSTGRES_IMAGE;
-  const password = process.env.DEV_INSTANCE_POSTGRES_PASSWORD || "qm-dev";
-  const volume = process.env.DEV_INSTANCE_POSTGRES_VOLUME || "qm-dev-postgres-data";
+export async function ensureLocalPostgres(
+  worktree: string,
+  env: Record<string, string>,
+  log: (msg: string) => void,
+): Promise<LocalPostgres> {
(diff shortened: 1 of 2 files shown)


ReganBell:

→ #221

PR #75, opened 2026-08-01
Custom MCP connectors

Luke20w:
## Summary
- Propose custom MCP connectors so the agent can use tools beyond the prebuilt SaaS connectors (Paper, ClickUp, etc.).

## Test plan
- [ ] Maintainers review the ADR in `adrs/custom-mcp-connectors.md` and align on scope before any implementation.

Made with [Cursor](https://cursor.com)

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/75"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788152259&installation_model_id=19911&pr_number=75&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F75&signature=3db25eac1348505b2208e522c9912f1663fb6b7dab8c6411b9153ce89f1cde7d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #172, opened 2026-08-03
adr: Granola connector

rsindall:
Requesting a first-class Granola connector (meeting notes) alongside the existing OAuth connectors.

What we'd want, roughly:
- users connect their own Granola account from Keychain
- agent can search / read meeting notes and transcripts in-scope for that user
- ideally read-only to start


ReganBell:
Thanks — appreciate the concise ask. Consolidating connector requests that reduce to "attach an MCP-shaped tool surface" under #75 (generic MCP connectors); Granola would ride that as a recipe. Closing in favor of #75.

PR #173, opened 2026-08-03
adr: Salesforce MCP connector

rsindall:
Requesting first-class Salesforce support for agents, preferably MCP-shaped (or whatever fits the existing connector model).


ReganBell:
Thanks — this is a well-shaped ask. Salesforce is exactly the kind of tool a generic MCP connector slot should cover, and we are consolidating that direction under #75. A Salesforce recipe on top of it would be the natural follow-on. Closing in favor of #75 — please add your requirements there.

PR #174, opened 2026-08-03
adr: GBrain connector

rsindall:
Requesting a first-class company knowledge-graph / GBrain connector (or MCP-shaped recipe).


ReganBell:
Thanks for the write-up. The knowledge-store angle is interesting and slightly different from the pure tool-surface asks, but the nearest-term path is still the generic MCP connector slot being consolidated under #75 — a GBrain recipe there gets you most of this. Closing in favor of #75; if the deeper "external brain" hook still feels needed after that lands, a fresh ADR scoped to just that would be welcome.

PR #229, opened 2026-08-05
Add the Claude Managed Agents harness

klesse413:
we'd like to add [claude managed agents](https://platform.claude.com/docs/en/managed-agents/overview) (CMA, anthropic's sessions API for server-hosted agents) as a harness option.

given this is a meaty change, proposing the code changes directly and including what i would have otherwise included in adr in pr description.

why add another claude harness
qm runs on a brain/hands split: the model reasons somewhere, and tools run on the user's own durable sandbox. every harness so far is a local loop that wasn't designed for that split, so qm makes it work by wrapping the loop and swapping its tools for bridged ones. CMA's harness outside sandbox architecture fits qm's architecture natively. the brain is a hosted session on anthropic's servers, the hands are the user's sandbox, and tool calls come back to qm to run there, with the user's credentials, under the same command policy and approvals as the other harnesses. Claude is able to use its native bash tool, and it executes on the user's sandbox through the same bridged execute path.

advantages of cma
reliability comes from the split architecture itself. the conversation lives in the session store on anthropic's side and tools run in qm's sandbox, so there's no in-process loop to lose: a redeploy of qm mid-task can't drop the conversation, and a turn sends just one new message instead of the whole history. performance comes from the model working with its own tools: claude uses its own bash tool rather than a stand-in like the claude code harness needs to use. the adapter is just an http client, so there's no harness runtime to operate. and every session has a live trace in the claude console, which makes debugging an agent much easier.

cmaturndiagram

tradeoffs
the API is beta. changing the model, effort, or system prompt rotates the session, reseeded from qm's log so nothing is lost. steering queues between turns. and the conversation exists on both sides, since qm's log is the source of truth.

testing & verification
tests mock the CMA API at the http boundary and cover turn flow, resume and rotation, strict gate on both tool routes, work-queue lease, aborts, and retries. qm check verifies the environment and its key against the live API before any deploy.

Screenshot 2026-08-04 at 2 34 52 PM

Screenshot 2026-08-04 at 2 54 17 PM

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/229"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788500208&installation_model_id=19911&pr_number=229&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F229&signature=7f9b902cbef1dfc4242cffd867f41e13043b884e1529fac63343fea0c29061af"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


src/harness/claude-harness.ts
```diff
@@ -188,63 +164,8 @@ class MessageQueue implements AsyncIterable<SDKUserMessage> {
   }
 }
 
-function toolOptions(opts: ClaudeHarnessOptions, turn?: HarnessTurnInput): PiToolsOptions {
```
klesse413, line 191: felt not great to add a 4th exact copy, so refactored all of them to use the pi copy. but would be happy to remove this change if the copies are intentional

Sibigi:
Nehi chahiye
Stop

time-attack:
Thanks for this, @klesse413 — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/743 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #231, opened 2026-08-05
Add BlueNexus integration: 200+ connections via single OAuth grant

tahpot:
## What this adds

This PR adds BlueNexus to the connector registry so a QM user can link it from the Keychain page, and adds a seeded skill so the agent knows how to use it. The full, authoritative list of supported connectors lives at https://bluenexus.ai/connectors.

Users can connect all their accounts — Slack, Notion, GitHub, Google Workspace, Telegram and 200+ others, with several thousand tools between them — behind a single OAuth grant. A QM user who links it gets whichever of those services they've already connected.

Users can add custom MCP servers to their BlueNexus account, providing a quick way to support any MCP server until QM natively supports it (see https://github.com/yc-software/qm/pull/75).

Connections can be permissioned as read-only or read + write, which controls what the agent is able to do once connected. This connector defaults to read + write (`universal-mcp-read-write`), exposing all six BlueNexus tools. To run it read-only instead, change `scopes` to `["universal-mcp-read"]` and register the BlueNexus app with that same scope — the two have to match, because BlueNexus rejects an authorization request asking for a scope the app does not allow.

I note the preference in CONTRIBUTING.MD for human `feature` PR's, but since we know our own API it seemed to make sense to do a traditional PR :)

## Demo

**▶️ Demo video:**

https://www.loom.com/share/356ffacf7577440884d3cbf53298dafe

A short walkthrough: Keychain → Connect BlueNexus → consent → the agent listing and calling downstream tools from a single grant.

## How it works

The model gets a small fixed tool surface and everything else runs through `execute` in the scope's sandbox. This connector does not change that. It follows the same shape as the existing Google, Dropbox and Linear connectors: core holds the credential, and a skill teaches the call.

1. Core stores the OAuth token in the keychain, encrypted with AES-256-GCM.
2. Each DM turn, core injects it into the sandbox as `$VAULT_TOKEN_BLUENEXUS_AI`, the same way `$VAULT_TOKEN_WWW_GOOGLEAPIS_COM` already works.
3. `skills-seed/bluenexus-connections/SKILL.md` tells the agent the endpoint, the JSON-RPC shape, the six tools, and the failure modes.
4. The agent calls `https://api.bluenexus.ai/mcp` with `curl`.

**Note: No tools are added to the model's context.** Several hundred downstream tools stay on the far side of one `execute` call, maximizing context window size for actual meaningful work.

## Test changes

Two existing tests needed updating, both because the specification genuinely changed rather than to make them pass.

`only X opts into PKCE (regression guard)` asserted exactly one PKCE provider. BlueNexus is the second deliberate opt-in, so the guard now takes an explicit allowlist of `x` and `bluenexus`. The `else` branch is untouched, so a third accidental opt-in still fails.

`the catalog endpoint exposes per-provider setup guidance` pins the sorted provider list, which gained `bluenexus`.

One test was added: `BlueNexus authorizes on the app host but exchanges on the api host, with PKCE and the read-write MCP scope`. That pins the split-host quirk — authorize on `app.bluenexus.ai`, token exchange on `api.bluenexus.ai` — which is unique among the eight providers and the easiest thing to "tidy" into a single host and silently break.

## Verification

`tsc` (root, cli, contract), `prettier`, `eslint`, `oxlint` and `knip` are clean.

CLI 503/503, web-ui 440/440, deployment stack contracts 1/1. Core is 3579/3713 with two failures in `opencode-harness.test.ts` and `sandbox-noninteractive.test.ts`; both were confirmed pre-existing by running them on unmodified `7f2c916`, and neither file touches anything here.

Exercised end to end against production BlueNexus from a cold start: Keychain → Connect account → consent → callback → grant stored with a refresh token → the agent listing all six tools using the injected env var.

## Setup

Create an OAuth client at BlueNexus → Developers → My Apps → Add App, in Third-Party Integration mode, grant it the same scope this connector requests (`universal-mcp-read-write` by default), and add `<PUBLIC_URL>/v1/connectors/oauth/bluenexus/callback` as its redirect URI. Clients can also be registered over the API; the `registration_endpoint` in `https://api.bluenexus.ai/.well-known/oauth-authorization-server` accepts RFC 7591 requests.

Then set the two secrets and restart:

```bash
BLUENEXUS_OAUTH_CLIENT_ID=client_...
BLUENEXUS_OAUTH_CLIENT_SECRET=...
```

`qm check` now fails when the client ID is set without the secret.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/231"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788501076&installation_model_id=19911&pr_number=231&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F231&signature=56de40bc612ceb14bf9d9adedcbe7d0c5460ae02848c168fbe8d3525762d43ef"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/cli/src/commands/setup.ts
+++ b/cli/src/commands/setup.ts
@@ -111,6 +111,7 @@ const PLAYBOOKS: Readonly<Record<string, readonly string[]>> = {
   GOOGLE_OAUTH_CLIENT_SECRET: ["From your Google Cloud OAuth client (APIs & Services -> Credentials)."],
   DROPBOX_OAUTH_CLIENT_SECRET: ["From your Dropbox app console (https://www.dropbox.com/developers/apps)."],
   LINEAR_OAUTH_CLIENT_SECRET: ["From your Linear OAuth application settings."],
+  BLUENEXUS_OAUTH_CLIENT_SECRET: ["From your BlueNexus OAuth client (Developer -> My Apps)."],
 };
 
 const FORMAT_HINTS: Readonly<Record<string, { prefix: string; label: string }>> = {
--- a/cli/src/secrets.ts
+++ b/cli/src/secrets.ts
@@ -164,6 +164,12 @@ export const FIRST_PARTY_SECRET_SPECS: readonly SecretSpec[] = [
     description: "Linear OAuth client secret.",
   },
   {
+    name: "BLUENEXUS_OAUTH_CLIENT_SECRET",
+    service: "core",
+    required: { when: { kind: "env-present", service: "core", name: "BLUENEXUS_OAUTH_CLIENT_ID" } },
+    description: "BlueNexus OAuth client secret.",
+  },
+  {
     name: "SLACK_BOT_TOKEN",
     service: "slack",
     required: false,
--- a/deploy/stacks/acme/.env.example
+++ b/deploy/stacks/acme/.env.example
@@ -77,6 +77,10 @@ SKILL_SIGNING_SECRET=
 # Needed when env.core.LINEAR_OAUTH_CLIENT_ID is set.
 # LINEAR_OAUTH_CLIENT_SECRET=
 
+# BlueNexus OAuth client secret. (core)
+# Needed when env.core.BLUENEXUS_OAUTH_CLIENT_ID is set.
+# BLUENEXUS_OAUTH_CLIENT_SECRET=
+
 # Slack request-signing secret (from the Slack app's Basic Information page); HTTP events mode only. (slack)
 # Needed when env.slack.SLACK_EVENTS_MODE is "http".
 # SLACK_SIGNING_SECRET=
--- a/plugins/web-ui/src/connector-link.ts
+++ b/plugins/web-ui/src/connector-link.ts
@@ -9,6 +9,7 @@ export const CONNECTOR_NAMES: Record<string, string> = {
   github: "GitHub",
   dropbox: "Dropbox",
   x: "X",
+  bluenexus: "BlueNexus",
 };
 
 export interface ConnectorLink {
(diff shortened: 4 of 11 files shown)


odfalik:
read https://github.com/yc-software/qm/blob/main/CONTRIBUTING.md

tahpot:
> read https://github.com/yc-software/qm/blob/main/CONTRIBUTING.md

Thanks @odfalik 

I mention that in the PR description.

> I note the preference in CONTRIBUTING.MD for human feature PR's, but since we know our own API it seemed to make sense to do a traditional PR :)

If you still prefer a human written, happy to switch it up.

time-attack:
Thanks for this, @tahpot — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/810 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #168, opened 2026-08-03
reach: accept threadTs to reply inside a channel or group thread

ReganBell:
POST /v1/reach on channel and group destinations now accepts an optional `threadTs`, so a caller can reply inside an existing Slack thread instead of always posting a new top-level message.

- `threadTs` is validated as a Slack timestamp and applies to `channel` and `participants` posts
- rejected with `recipient` (person DMs have no channel threads) and with `react`/`delete`, which already name their target `ts`
- `reachNow` threads the resolved slack/group destination via a `withThread` helper alongside `withReact`/`withDelete`
- the API catalog documents the parameter; tests cover accept and reject paths

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/168"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788390181&installation_model_id=19911&pr_number=168&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F168&signature=fda542eabe749b8a4e4c5b8b6dcf82894b596dc5f3241afe959f1d8a3239be71"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #235, opened 2026-08-05
adrs: let assistant replies thread inside a 1:1 DM

odfalik:
Just a short note in `adrs/` asking whether assistant replies can thread inside a 1:1 DM. Slack allows it, but a thread id on a person-DM currently comes back rejected.


odfalik:
https://github.com/yc-software/qm/pull/168

PR #242, opened 2026-08-06
auth: prefer AUTH LOGIN over PLAIN when the server advertises both

LeoneNee:
## Problem

`authSteps` prefers AUTH PLAIN whenever the server advertises it. Some servers list PLAIN in their EHLO capabilities but reject it in practice — observed with QQ Mail (`smtp.qq.com`), which answers an `AUTH PLAIN` attempt with `502 Invalid input` / `535 Login fail` yet completes `AUTH LOGIN` with the same credentials. Sign-in email delivery fails with valid credentials and no recourse, because PLAIN is always picked when advertised.

## Change

Try AUTH LOGIN first; fall back to PLAIN only when LOGIN is not advertised. LOGIN is a single extra round trip and is accepted everywhere PLAIN is, so preferring it strictly widens server compatibility.

## Tests

- The fake SMTP server in `smtp.test.ts` now speaks the LOGIN challenge-response (`334` base64 prompts) instead of treating every `AUTH` line as one-shot.
- The delivery and verification tests assert the LOGIN transcript (`AUTH LOGIN`, base64 username, base64 password).
- A new case covers a server advertising only PLAIN, keeping the fallback path covered.

All 8 tests in `plugins/auth/test/smtp.test.ts` pass. The fix was also verified against a live QQ Mail account: `AUTH LOGIN` completes (`235 Authentication successful`) where `AUTH PLAIN` is rejected.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/242"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788592817&installation_model_id=19911&pr_number=242&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F242&signature=0978336f361d7d3d59e6cc6de304fad74489ac6c15b2642149e2b6379ded2a10"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/auth/src/smtp.ts
+++ b/plugins/auth/src/smtp.ts
@@ -159,13 +159,13 @@ async function openSocket(options: SmtpOptions, timeoutMs: number): Promise<Sock
 }
 
 function authSteps(mechanisms: string, username: string, password: string): string[] {
-  if (/\bPLAIN\b/.test(mechanisms))
-    return [`AUTH PLAIN ${Buffer.from(`\0${username}\0${password}`, "utf8").toString("base64")}`];
-  return [
-    "AUTH LOGIN",
-    Buffer.from(username, "utf8").toString("base64"),
-    Buffer.from(password, "utf8").toString("base64"),
-  ];
+  if (/\bLOGIN\b/i.test(mechanisms))
+    return [
+      "AUTH LOGIN",
+      Buffer.from(username, "utf8").toString("base64"),
+      Buffer.from(password, "utf8").toString("base64"),
+    ];
+  return [`AUTH PLAIN ${Buffer.from(`\0${username}\0${password}`, "utf8").toString("base64")}`];
 }
 
 function clientName(host: string): string {
--- a/plugins/auth/test/smtp.test.ts
+++ b/plugins/auth/test/smtp.test.ts
@@ -11,7 +11,7 @@ interface FakeSmtp {
 }
 
 async function fakeSmtp(
-  options: { offerStartTls?: boolean; rejectAuth?: boolean; rejectRecipient?: boolean } = {},
+  options: { offerStartTls?: boolean; rejectAuth?: boolean; rejectRecipient?: boolean; mechanisms?: string } = {},
 ): Promise<FakeSmtp> {
   const transcript: string[] = [];
   const messages: string[] = [];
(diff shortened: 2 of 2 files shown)


Brown226:
Found a counterpart that this fix leaves inconsistent: `cli/src/preflight.ts` `smtpVerify` (around line 250) still prefers `AUTH PLAIN` whenever the server advertises it, the exact inverse of the new delivery-side logic in `plugins/auth/src/smtp.ts`.

Concrete failure mode: a server that advertises both PLAIN and LOGIN but rejects PLAIN on use (the QQ Mail case described here) now delivers mail fine — but `qm check`/`qm doctor` preflight will still report SMTP auth as failed, because preflight uses PLAIN while delivery uses LOGIN. The operator sees a red preflight for a transport that actually works.

Proposal: apply the same preference (`LOGIN` if advertised, else `PLAIN`) to `smtpVerify`'s step selection, and keep the existing `expect` codes (235/334) as-is — the LOGIN challenge-response already works through the same loop.

Tested in a private fork: with the delivery-side change alone, smtp.test.ts passes 8/8, but the preflight inconsistency is visible by inspection. Happy to submit the preflight-side change as a follow-up PR if useful.

time-attack:
Thanks for this, @LeoneNee — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/808 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #243, opened 2026-08-06
local-sandbox: allow reaching the exec daemon over a peer network

LeoneNee:
## Problem

`local-sandbox` reaches the sandbox exec daemon at a hard-coded `http://127.0.0.1:<port>`. That only works when core runs on the docker host itself. When core runs inside a container — the shape `deploy/core/Dockerfile` produces — loopback inside the core container is the container itself, not the host publishing the sandbox's mapped agent port. Every exec/read/write against the sandbox fails with `fetch failed`, and the agent reports its computer backend down.

## Change

With `LOCAL_SANDBOX_PEER_NETWORK` set:

- sandbox containers are attached to that docker network after `docker run` / `docker start` (`docker network connect`, idempotent)
- core reaches the daemon by container name on `AGENT_PORT` instead of via the published loopback port

Unset keeps the loopback behavior used when core runs on the host, so existing single-process setups are unaffected.

The exec daemon is unauthenticated (loopback binding is its access control), so this option is only for a trusted internal network — e.g. the private stack network shared with the core container, never a public one.

## Tests

`test/local-sandbox.test.ts` passes (15/15). The option was also exercised end to end with core in a container: a turn executed in the sandbox container and a file written through the exec daemon was verified from the host volume.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/243"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788593142&installation_model_id=19911&pr_number=243&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F243&signature=9080298a1ce1f1a2aa34890ccbbe59e993c45763ac6bc24537b29beaccad4d15"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -245,12 +245,14 @@ interface LocalSandboxEnv {
   cpus?: number;
   memoryMb?: number;
   defaultTimeoutSec?: number;
+  peerNetwork?: string;
 }
 
 function localSandboxEnv(env: NodeJS.ProcessEnv): LocalSandboxEnv {
   return {
     ...(env.LOCAL_SANDBOX_IMAGE ? { image: env.LOCAL_SANDBOX_IMAGE } : {}),
     ...(env.LOCAL_SANDBOX_DOCKER_BIN ? { dockerBin: env.LOCAL_SANDBOX_DOCKER_BIN } : {}),
+    ...(env.LOCAL_SANDBOX_PEER_NETWORK ? { peerNetwork: env.LOCAL_SANDBOX_PEER_NETWORK } : {}),
     ...(numEnvStrict("LOCAL_SANDBOX_CPUS", env.LOCAL_SANDBOX_CPUS) !== undefined
       ? { cpus: numEnvStrict("LOCAL_SANDBOX_CPUS", env.LOCAL_SANDBOX_CPUS) }
       : {}),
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this, @LeoneNee — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/829 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

Issue #248, opened 2026-08-06
chassis cookie(): malformed percent-encoding in a cookie value returns 502/500 instead of signed-out


joeVenner:
A cookie value with malformed percent-encoding makes `cookie()` in `plugins/chassis/src/http.ts` throw, so a request that should read as "signed out" comes back as a server error instead.

Affected revision: `0f0e0ad`.

## What happens

`cookie()` passes the matched value straight to `decodeURIComponent`:

```ts
export function cookie(req: IncomingMessage, name: string): string | null {
  const m = (req.headers.cookie ?? "").match(new RegExp(`(?:^|;\\s*)${name}=([^;]+)`));
  return m ? decodeURIComponent(m[1] ?? "") || null : null;
}
```

`decodeURIComponent` throws `URIError: URI malformed` on an invalid percent-escape. The throw escapes the helper and is caught only by each surface's top-level handler wrapper, so:

- web-ui — `authenticate()` (`plugins/web-ui/server/index.ts:263`) throws, the wrapper at `server/index.ts:1919` logs and returns **502 `bad_gateway`** where the request should have been a 401 `sign in`.
- admin — `cookie(req, "admin")` (`plugins/admin/src/index.ts:89`) throws, and the wrapper at `src/index.ts:298` returns **500 `internal_error`**.

Both wrappers do catch it, so the process does not crash and nothing fails open — it's a wrong status code, not an auth bypass. The annoyance is that the browser resends the bad cookie on every subsequent request, so the surface stays broken for that client until the cookie is cleared by hand.

## Repro

```
$ curl -s -o /dev/null -w '%{http_code}\n' -H 'Cookie: webuiuser=alice' http://localhost:PORT/me
200
$ curl -s -o /dev/null -w '%{http_code}\n' -H 'Cookie: webuiuser=%' http://localhost:PORT/me
502
```

`webuiuser=%`, `webuiuser=%E0%A4%A`, and `webuiuser=alice%zz` all reproduce; `webuiuser=100%25` (valid encoding) does not.

There is a second case the regex causes on its own: it takes the *first* match, so a malformed duplicate shadows a valid cookie of the same name — `Cookie: webuiuser=%; webuiuser=alice` fails even though a usable session cookie is present.

## Why this looks like an oversight rather than intent

Every other cookie/component reader in the repo already guards the decode:

- `plugins/portal/src/session.ts:130` — `readCookie` wraps it in try/catch (and regex-escapes the name)
[...]

PR #249, opened 2026-08-06
ci: verify documented npm commands exist (and drop two that never did)

joeVenner:
`fly/README.md` documents two commands that do not exist:

```bash
FLY_API_TOKEN="$(fly tokens create deploy -a "$FLY_SANDBOX_APP_NAME")" npm run smoke:fly
FLY_API_TOKEN=... npm run smoke:x
```

Neither `smoke:fly` nor `smoke:x` is defined in any `package.json`, and `git log -S` shows neither has ever been defined in this repository's history. There is no `scripts/fly-smoke.ts` either, so the paragraph describing what "the smoke test" backs up and restores documents a run nobody can perform. The `X_SMOKE_REQUIRE_AUTH` and `X_SMOKE_REQUIRE_FIREHOSE` env vars in the same block appear nowhere outside that file.

The interesting part isn't the two stale blocks — it's that nothing would have caught them. Every other class of rot in this repo has a gate: `knip` for dead code, `oxlint` and `eslint` for source, `prettier` for formatting. Documented commands had none, so these sat indefinitely.

So this PR is mostly the guard, with the two fixes that make it pass.

## Commits

**`docs(fly): drop smoke commands that were never implemented`** — removes the two command blocks and the prose that exists only to describe them. Removed rather than corrected because there is no correct name to point at; if those scripts are pending, the right move is to re-add the docs with them and I'm happy to drop this commit. The `npm run smoke:git-cli` block below it is real and is left untouched.

**`ci: verify documented npm commands exist`** — adds `scripts/check-doc-npm-scripts.mjs`, a `lint:docs` script, and one step in the `lint` job next to `lint:knip` / `lint:ox`. It reads every tracked `*.md`, extracts each `npm run <script>`, and fails if no tracked `package.json` defines that script.

## The one design decision

Resolution is a **union across all packages**, not the package nearest the file. Docs legitimately show commands to run from another directory — root `README.md` documents `npm run build` and `npm run serve`, which only `plugins/web-ui` and `cli` define — so matching per directory would reject correct docs.

The tradeoff: the union catches the case that matters (a command no package defines anywhere) but cannot catch a command documented against the wrong directory. I went with the version that has no false positives, on the grounds that a lint gate which fires on correct docs gets ignored or removed. Happy to tighten it if you'd rather have the stricter check plus an allowlist.

## Verification

On this branch the checker reports:

```
Checked 26 documented npm command(s) (14 distinct) across 7 packages — all defined.
```

Against `main` before the docs commit, it reports exactly the two real hits and nothing else — no false positives across the whole repo, including `.claude/skills/*.md`, `cli/`, `fly/`, `deploy/`, and the plugin READMEs:

```
  fly/README.md:181: npm run smoke:fly
  fly/README.md:187: npm run smoke:x
```

Negative-tested by appending `npm run definitely-not-a-script` to `deployment.md` — caught with file:line, exit 1.

Full lint job run locally and green: `format:check`, `lint`, `lint:knip`, `lint:ox`, `lint:docs`. Commits are ordered so each is green on its own — the docs fix lands before the gate that would flag it.

Nothing outside the lint job is touched, so no runtime, test, or build behavior changes.

## Note on process

`CONTRIBUTING.md` asks for bugs as issues rather than PRs, and I want to be upfront that I'm aware of that. I went with a PR here because the substance is a CI script rather than a product change, and describing it in prose seemed strictly worse than showing it. If you'd prefer this as an issue, or as an `adrs/` note, say the word and I'll move it.

Separately, I filed #248 for an unrelated bug in `plugins/chassis/src/http.ts` following the issue path.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/249"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788641027&installation_model_id=19911&pr_number=249&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F249&signature=b4f34fee8565372958a3f9d4c7241f7591d4cf23ea4c171e5c9db8cb97e4cd46"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -146,6 +146,8 @@ jobs:
         run: npm run lint:knip
       - name: Oxlint
         run: npm run lint:ox
+      - name: Documented npm commands exist
+        run: npm run lint:docs
 
   core-postgres:
     name: Core Postgres tests
--- a/fly/README.md
+++ b/fly/README.md
@@ -177,27 +177,6 @@ For resident X tooling use the same prefix for native tool env, for example
 
 ## Smoke test
 
-```bash
-FLY_API_TOKEN="$(fly tokens create deploy -a "$FLY_SANDBOX_APP_NAME")" npm run smoke:fly
-```
-
-For image-resident X helper readiness:
-
-```bash
-FLY_API_TOKEN=... npm run smoke:x
-```
-
-This verifies `x-api` is on PATH and reports `missing_auth=auth_missing` when no
-resident X token is installed. Add `X_SMOKE_REQUIRE_AUTH=1` after configuring
-`X_BEARER_TOKEN` / `X_ACCESS_TOKEN`, or `X_SMOKE_REQUIRE_FIREHOSE=1` when a vendored
-`x-firehose` binary should be present.
-
-The smoke test uses a timestamped personal smoke-test scope, writes
-workspace and resident-home state, backs it up, deletes the Fly machine, recreates it,
-restores the backup, verifies `.aws/*` stayed excluded, then deletes the smoke
-machine. Add `SNAPSHOT_STORE=s3 S3_BUCKET=...` to exercise the real S3 object store;
-without those vars it uses the same backup-store code over an in-memory blob store.
-
 For GitHub/GitLab resident CLI readiness:
 
 ```bash
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
     "lint": "eslint .",
     "lint:ox": "oxlint --deny-warnings src plugins scripts cli test",
     "dev-instance:doctor": "bash scripts/dev-instance.sh doctor",
-    "lint:knip": "knip"
+    "lint:knip": "knip",
+    "lint:docs": "node scripts/check-doc-npm-scripts.mjs"
   },
   "dependencies": {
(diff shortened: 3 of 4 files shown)


time-attack:
Thanks for this, @joeVenner — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/807 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #254, opened 2026-08-07
feat(web-ui): surface delivered posts in live replies

ErciZe:
## Problem

Web runs can deliver their final answer through a successful `post` tool call and then terminate without `result.reply`. The live transcript currently renders only the work timeline until the session is reloaded, even though history reconstruction already treats the delivered post as the reply.

## Change

- Promote the latest successfully delivered post to the terminal live reply when no ordinary reply exists.
- Preserve failure and abort semantics while retaining text that was already delivered.
- Replace stale reattach text when the delivered post is not an append-only continuation.

## Verification

- Web UI tests: 502 passed
- Web UI typecheck
- Web UI production build
- Prettier, ESLint, and Oxlint on changed files
- Fresh-context review: no findings

## Screenshot

Synthetic data, rendered without deployment-specific content:

Before and after: delivered post appears in the live transcript

### DIFF
--- a/plugins/web-ui/src/core-bridge.ts
+++ b/plugins/web-ui/src/core-bridge.ts
@@ -854,8 +854,10 @@ function applyRun(
     if (paused) work.pendingApprovals = approvals;
     notify?.();
   }
+  const deliveredPost = latestDeliveredPostText(work?.activity ?? []);
   if (res?.stopped) {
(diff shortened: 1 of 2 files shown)


ErciZe:
Closing this contribution; the change will remain deployment-local for now.

PR #258, opened 2026-08-07
ADR: what blocked a first AWS deployment (August 2026)

wooheum-xin:
Notes from standing up a fresh `--target aws` deployment, following the process in `CONTRIBUTING.md` — a `.md` file in `adrs/`, informal, describing what we hit rather than proposing an implementation.

Seven things stopped the deployment until each was worked around. Roughly by how much damage they do:

1. `ADMIN_GRANTS` is a first-boot seed, not a setting. Seeding an address that cannot sign in locks the deployment out of its own Admin area, and recovery needed a one-off ECS task writing to `admin_grants` directly because RDS is VPC-private and ECS Exec is off.
2. The `gh-2.96.0-1` pin in the MicroVM agent image has rotted — the RPM repo drops old versions, so `qm infra build-image` fails. AWS surfaces no diagnostic at all: `CREATE_FAILED` with no `stateReason` and a log group with no stream.
3. Published images are amd64 only while the scaffold defaults `architecture` to `arm64`.
4. The `npm audit` gate in `deploy/core/Dockerfile` blocks `--build-from` when a production advisory has no fix available.
5. Running the CLI from a source checkout hits the `registry.invalid` sentinel and fails deep into the deploy, after infrastructure exists and secrets are pushed.
6. Quoted values in `.env` become part of the secret and travel to the provider.
7. The agent misdiagnoses missing credentials in shared scopes, pointing users at an administrator setting that does not exist instead of the keychain grant flow.

Plus smaller notes on `infra delete-image`, secret-only changes not restarting tasks, and the public-URL check assuming CloudFront when a proxying CDN sits in front of the ALB.

Happy to expand on any of these, and glad to test fixes against a live AWS deployment.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/258"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788676448&installation_model_id=19911&pr_number=258&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F258&signature=daf6201550dc9d0f8e5cd21660e5de2b1301fe45176086f6ff17afe8536c43b6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/adrs/aws-deployment-blockers-2026-08.md
+++ b/adrs/aws-deployment-blockers-2026-08.md
@@ -1,0 +1,101 @@
+# Things that blocked a first AWS deployment (August 2026)
+
+Notes from standing up a fresh `--target aws` deployment (ap-northeast-1, Slack OIDC,
+in-house model endpoint). Each item stopped the deployment until it was worked around.
+Listed roughly by how much damage it does.
+
+## 1. `ADMIN_GRANTS` is a first-boot seed, and nothing says so
+
+`bootAdminGrantSeed` is only applied when the grant table is empty
+(`src/admin/admin-grant-store.ts`). Once the stack has booted, editing `ADMIN_GRANTS`
+and redeploying silently does nothing.
+
+Seeding an address that cannot sign in locks the deployment out of its own Admin area:
+the seeded admin can never authenticate, and the person who can authenticate cannot be
+granted anything, because granting requires an existing admin. We hit exactly this by
+seeding a shared mailbox that had no Slack account under Slack OIDC.
+
+Recovery required running a one-off ECS task on the core task definition to insert a row
+into `admin_grants` directly, because RDS is VPC-private and ECS Exec is off. That is not
+a procedure an operator can be expected to invent.
+
+Worth having: `qm check` or `qm doctor` warning when `ADMIN_GRANTS` differs from the live
+durable grants, a documented recovery path, and a note in `.env.example` that the value is
+a seed rather than a setting.
+
+## 2. Pinned `gh` version in the MicroVM agent image has rotted
+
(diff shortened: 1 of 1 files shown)


wooheum-xin:
One follow-up on the `gh` pin, since I could check it directly.

The RPM repository at `cli.github.com/packages/rpm` currently carries exactly one `gh` version — `2.97.0-1`. `2.96.0-1` is gone from the metadata entirely, so the pin in `aws/microvm-agent/Dockerfile` and its copy under `cli/templates/` cannot resolve today.

Worth noting that `cli/test/microvm-dockerfile.test.ts` asserts a pin is present (`/dnf install -y gh-\d+\.\d+\.\d+-\d+/`), so dropping the pin is a test change too, not just a Dockerfile one. Either direction works from our side — we bumped to `2.97.0-1` locally to get building again, which is the smaller change but will rot the same way the next time the repo rolls forward.

PR #299, opened 2026-08-09
adrs: FundOS fund-operations skills for QM

rchachra:
Hi,

Vijay and I are the founders of VantedgeAI (W22). We build FundOS, an operating system
for VC, PE and private credit funds covering compliance and operations for LP records,
capital calls, NAV, waterfalls, covenants and data rooms.

We have been testing QM against FundOS and have written seven skills that let an agent do
actual fund operations: draft a compliant capital call from the live LP roster, run a
distribution waterfall and check portfolio covenants.

The skills can be found here:
https://github.com/8vdx1/fundos-mcp/tree/main/integrations/qm

Two points worth noting:

1. We initially assumed we could point QM at our MCP server. It took reading
   claude-harness.ts to understand that the harness uses MCP as its own internal
   transport. That makes sense, but the SDK sitting in devDependencies sent us down the
   wrong path. Maybe it makes sense to add a note in the README as it may save others the
   same detour.

2. For us, the credential broker is the most valuable part. Some of these workflows can
   affect real LP money, so having the agent never hold the credentials matters far more
   than convenience. Maybe this feature should be highlighted more prominently.

We would be happy to PR the skills into skills-seed/ if they are useful. Otherwise we can
leave them in our repo. Let us know whether a vertical-specific fund operations pack makes
sense for the seed set.

Best,
Ravi

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/299"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788845029&installation_model_id=19911&pr_number=299&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F299&signature=6050e01a839ac18e124c780c7f4128bb549558257340f5c8c5c0a799f445ae67"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/adrs/fundos-fund-operations-skills.md
+++ b/adrs/fundos-fund-operations-skills.md
@@ -1,0 +1,31 @@
+Hi,
+
+Vijay and I are the founders of VantedgeAI (W22). We build FundOS, an operating system
+for VC, PE and private credit funds covering compliance and operations for LP records,
+capital calls, NAV, waterfalls, covenants and data rooms.
+
+We have been testing QM against FundOS and have written seven skills that let an agent do
+actual fund operations: draft a compliant capital call from the live LP roster, run a
+distribution waterfall and check portfolio covenants.
+
+The skills can be found here:
+https://github.com/8vdx1/fundos-mcp/tree/main/integrations/qm
+
+Two points worth noting:
+
+1. We initially assumed we could point QM at our MCP server. It took reading
+   claude-harness.ts to understand that the harness uses MCP as its own internal
+   transport. That makes sense, but the SDK sitting in devDependencies sent us down the
+   wrong path. Maybe it makes sense to add a note in the README as it may save others the
+   same detour.
+
+2. For us, the credential broker is the most valuable part. Some of these workflows can
+   affect real LP money, so having the agent never hold the credentials matters far more
+   than convenience. Maybe this feature should be highlighted more prominently.
+
+We would be happy to PR the skills into skills-seed/ if they are useful. Otherwise we can
+leave them in our repo. Let us know whether a vertical-specific fund operations pack makes
+sense for the seed set.
+
+Best,
+Ravi


rchachra:
One more note in case it's useful for evaluating these.

The skills are plain REST over `https://fundos.vantedgeai.com/api/v1` — no MCP, no SDK, no
build step. Each `SKILL.md` is readable on its own, so the quickest review is just opening
`fundos/SKILL.md` (the root one: auth, the shell helper, the safety contract) and then any
domain skill next to it. Every example is a literal curl you can eyeball.

Two things we deliberately put in there that you may or may not agree with:

- Credentials go through QM's broker rather than a token in the sandbox, so the agent never
  holds the key and your admin pins host/method/path at QM's edge.
- Writes are proposals. FundOS routes every write to a human approval queue server-side, and
  the skills are written to report "drafted for approval", never "issued" — a capital call
  moves real LP money and an agent should not be the last step.

If you want to actually run them rather than read them, tell me and I'll set you up with a
sandbox workspace and a credential — synthetic fund data, nothing real, so you can push it
around freely. Not posting one here for obvious reasons.

Happy either way — no need to run anything to tell us whether a vertical pack belongs in the
seed set.

ababove20-jpg (approved):

PR #303, opened 2026-08-09
web-ui: render custom-provider models instead of dropping them

hugo2046:
## Problem

Registering a custom model provider promises "the models join the picker" (the admin surface's Custom providers copy), and `/v1/surface-config` does advertise them — but the web-ui model picker never shows them.

## Root cause

`plugins/web-ui/src/pi-models.ts` `getBaseModel()` resolves unknown model ids through a fallback that only accepts `provider === "openrouter"`. Any other custom provider throws `Unsupported model`, and the `catch` in `model-options.ts` silently drops the entry. The file's own leading test is named "resolves models from the shared catalog **without privileging a provider**" — the implementation privileged one.

## Change

Two halves, 4 files:

- `src/api/routes/surface.ts` — surface-config's `modelCatalog` entries now carry the model's `api` protocol alongside `name`/`provider`. `resolveModel()` already resolves custom models and knows their protocol; the route just never shipped it.
- `plugins/web-ui/src/pi-models.ts` — the fallback clones a template picked by that `api` protocol (`anthropic-messages` → the anthropic template, otherwise the openai one; openrouter keeps its existing template) instead of gating on the provider id, and keeps the entry's own `provider`/`api` on the clone so downstream logic doesn't see the template's.
- `plugins/web-ui/src/model-options.ts` — catalog entry type gains the optional `api`.
- `plugins/web-ui/test/pi-models.test.ts` — regression test with a neutral custom provider for both protocols, plus the existing "unknown id still throws" case.

## Verification

- `npm run typecheck` clean
- `node --test plugins/web-ui/test/pi-models.test.ts plugins/web-ui/test/model-options.test.ts` — 18/18 pass
- Reproduced live before the fix: registered custom provider's models advertised by surface-config, absent from the picker; present with the fix.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/303"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788862266&installation_model_id=19911&pr_number=303&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F303&signature=273a3bab62f071f0156150cd306fde0572e9d3ad8a92fc8a97bf87b34820e00b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/web-ui/src/model-options.ts
+++ b/plugins/web-ui/src/model-options.ts
@@ -78,7 +78,7 @@ function buildOption(
   id: string,
   harnessId = "pi",
   qualified = false,
-  catalog: Readonly<Record<string, { name: string; provider: string }>> = {},
+  catalog: Readonly<Record<string, { name: string; provider: string; api?: string }>> = {},
 ): ModelOption | null {
   try {
     const dynamic = catalog[id];
@@ -101,7 +101,7 @@ function buildOptions(
   ids: readonly string[],
   harnessId = "pi",
   qualified = false,
-  catalog: Readonly<Record<string, { name: string; provider: string }>> = {},
+  catalog: Readonly<Record<string, { name: string; provider: string; api?: string }>> = {},
 ): ModelOption[] {
   const seen = new Set<string>();
   const out: ModelOption[] = [];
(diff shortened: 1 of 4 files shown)


hugo2046:
Superseded by #512 (merged 2026-08-13), which solves this a different way.

Instead of widening the known-provider list, `getBaseModel` now takes an optional `fallback: { name, provider }`; when a model id matches neither a builtin nor a clone template, it clones the `openrouter/auto` template and overrides `provider` with the custom one. `model-options.ts` passes the dynamic catalog entry (`getBaseModel(id, dynamic)`), so custom-provider models render instead of being dropped — which was the goal here.

Closing.

Issue #269, opened 2026-08-07
A configured custom-provider model is silently replaced by the built-in default at boot


ianTPE:
# A configured custom-provider model is silently replaced by the built-in default at boot

**Version:** CLI 0.1.4, repo at `0f0e0ad`

Related: #237 and #250 (the admin API can't register a custom provider) and #236 (the Web UI
picker drops custom models). This is a third facet of the same custom-provider wiring, and the
one that bites after you work around the other two: the provider is registered and the model id
is configured, and the deployment still runs a different model without saying so.

## What happens

With `HARNESS=pi` and `PI_MODEL=<a custom-provider model>`, turns run against
`claude-opus-5` instead. Nothing warns; the only visible symptom is a provider error, because
the deployment's key belongs to the custom provider, not to Anthropic:

```
Model provider API error (authentication_error): invalid x-api-key
```

`session_llm_requests` shows `"model":"claude-opus-5"` for a turn that should have used the
configured model.

## Why

Custom providers hydrate asynchronously and nothing waits for the initial pass:

```js
const refreshCustomProviders = async () => {
  setCustomProviders(await customProviders.enabled());
};
void refreshCustomProviders().catch(...);
```

Before that hydration finishes, `buildApp` constructs the actual router fallback in
`src/wiring.ts:772-780`:

```js
const fallback = {
  harnessId: fallbackHarness,
  modelId: defaultModelForHarness(
    fallbackHarness,
    configuredModelForHarness(config, fallbackHarness),
    baseModelProviders(config),
  ),
};
```

`src/index.ts` independently performs the same early computation for the HTTP/Admin
`baseModelDefault`. That copy affects the server and Admin view; it is not the router's
fallback.

At both call sites the registry is still empty, so `isCustomModelId` is false, `resolveModel`
returns undefined, `modelSupportedByHarness` returns false, and `defaultModelForHarness` falls
back to `DEFAULT_AGENT_MODEL_ID` (`src/model/pi-models.ts:8` — `claude-opus-5`).

The `fallback.modelId` created in `wiring.ts` is then captured for the life of the process.
`resolveRuntimeChoiceDurable` uses it whenever no durable scope row overrides it, and
`createHarnessRouter` forwards the result as `turn.model`. Hydrating the registry later does
not recompute that captured fallback.

## Effect

On a fresh process boot, `PI_MODEL` cannot name a custom-provider model: the race is
deterministic, not intermittent. The failure is silent — no `[swallowed]` line, no error event —
[...]

PR #312, opened 2026-08-10
fix: make custom providers usable end-to-end in the Web UI

### LINKED ISSUE #236: Custom provider models are missing from the Web UI model picker
Models from custom providers are returned correctly by `/api/runtime-config`, but they do not appear in the chat model picker. The picker falls back to the default built-in models.

## Reproduction

1. Configure a custom provider.
2. Add its models to the Web UI model allowlist.
3. Confirm the models appear in `modelsByHarness` and `modelCatalog` from `/api/runtime-config`.
4. Open the chat model picker.

## Cause

`plugins/web-ui/src/pi-models.ts#getBaseModel` only creates dynamic models when the provider is `openrouter`. Models from other providers throw `Unsupported model` and are silently dropped by `buildOption`.

## Expected

Models from any supported custom provider should appear in the picker.

The provider protocol could be included in `modelCatalog` so the frontend can construct the appropriate model instead of special-casing OpenRouter.

### LINKED ISSUE #282: Custom providers are 404 in production: src/index.ts never passes customProviders to createServer
Registering a custom model provider from the Admin UI fails with `Could not save this provider.` Every custom-provider route returns 404 — `GET /v1/admin/custom-providers`, `PUT .../:provider`, `DELETE .../:provider`.

The cause is not configuration. `src/wiring.ts` builds the store and returns it, `ServerDeps` in `src/api/deps.ts` declares `customProviders?: CustomProviderStore`, and the routes require `ctx.deps.customProviders`. But `src/index.ts` never passes it to `createServer` — it forwards `modelCredentials: built.modelCredentials` and stops there. Because the field is optional, omitting it typechecks, so `ctx.deps.customProviders` is `undefined` at runtime and `putCustomProvider`'s first guard returns `404 {"error":"not_found"}`.

Observed on a self-built image from `main`:

```
[admin] PUT /v1/admin/custom-providers/<slug> 404 27ms
```

`refreshCustomProviders` is missing the same way, so even a successful upsert would not re-hydrate the registry.

Why tests do not catch it: `test/custom-provider-route.test.ts` builds its own server and passes `customProviders: built.customProviders` and `refreshCustomProviders: built.refreshCustomProviders` explicitly. The test knows the correct wiring; only the production entrypoint does not. So the suite stays green while the feature is dead for every deployment.

The fix is two lines in `src/index.ts` next to `modelCredentials`:

```ts
customProviders: built.customProviders,
refreshCustomProviders: built.refreshCustomProviders,
```

We are running that locally and can confirm it makes the routes reachable. A regression test that asserts the entrypoint forwards them is worth having too, since the optional field means the next omission will also typecheck — we added one to `test/custom-provider-route.test.ts` and verified it fails when the fix is reverted.

Happy to send the patch if useful, though CONTRIBUTING suggests issues for bugs, so filing it here.

https://claude.ai/code/session_01AnKuftwBchrSLpfQkzBHUE

### LINKED ISSUE #290: Provider-key checks ignore the active harness — claude-harness instance locked out of its own web UI
I run qm with `HARNESS=claude` on a Claude subscription (`CLAUDE_CODE_OAUTH_TOKEN`, no `ANTHROPIC_API_KEY`). The assistant works — I send it a message and get a real reply.

But the browser can't reach the chat UI. `/` redirects to `/admin/onboarding`, which demands a model provider key, and there's no way past it. Storing *any* key gets me in — I used OpenRouter, for a provider this instance never calls. Non-admins get a 503 "not set up yet" page instead.

Once in, the onboarding page still says **"Needs a key — claude-opus-5 cannot run until its Anthropic key is configured"** about the model that just answered me.

Three places ask whether a provider key exists without asking whether the harness needs one:

- `src/api/routes/surface.ts:1041` — `modelProviderConfigured` from raw store availability; the portal redirects on it (`plugins/portal/src/index.ts:1056`).
- `plugins/admin/public/index.html:7050-7065` — status derived from the base model's provider.
- `plugins/admin/public/index.html:7047` — onboarding's model dropdown, unfiltered by harness; picking an OpenRouter model saves the key, then fails the base-model PUT with "Key saved, but the base model could not be changed." and no reason shown.

`modelProviderAvailabilityFor` already answers this correctly (`src/model/pi-models.ts:231`), and `runtimeConfigBody` forty lines below `surface.ts:1041` uses it. Codex may be the same shape via `CODEX_ACCESS_TOKEN`.

xieqian0816:
## Summary

This makes an admin-registered custom model provider usable across the complete production Web flow: save it, pass readiness checks, expose its model in the picker, and run Web turns without requiring a second, explicit allowlist configuration.

Closes #236
Closes #282

This also fixes the custom-provider instance of the readiness symptom described in #290. It does **not** close #290's broader Claude/Codex subscription-harness case, and it intentionally does not address the separate boot-time `PI_MODEL` hydration race in #269.

## Root cause and fix

| Broken boundary | Root cause | Fix |
| --- | --- | --- |
| Production Admin routes | `src/index.ts` did not pass `customProviders` or `refreshCustomProviders` into `createServer` | Wire both dependencies and add an entrypoint regression assertion |
| Portal/Admin readiness | Readiness only considered built-in credential availability | Include enabled, keyed custom-provider status without exposing credentials |
| Web turn validation | With no explicit picker allowlist and no OpenRouter key, validation fell back to built-in model IDs | Derive the default allowed set from the built-in + custom catalog while preserving explicit allowlists |
| Web model picker | Runtime config only supplied `name` and `provider`, so arbitrary custom models could not be materialized | Add protocol and token-window metadata, then construct a model from the correct protocol template |
| Admin refresh | Onboarding did not load the custom-provider table on entry and could retain stale readiness after a save/remove | Load the table with onboarding and refresh readiness after both mutations |

The implementation stays generic for both supported custom-provider protocols (`openai-completions` and `anthropic-messages`); it does not special-case DeepSeek.

## Security and compatibility

- API keys remain encrypted, write-only, and absent from every response, runtime payload, screenshot, and fixture.
- Runtime metadata adds only `api`, `contextWindow`, and `maxTokens`; the provider base URL and key are not sent to the browser.
- Catalog fields are additive and optional, preserving existing built-in and OpenRouter clients.
- An explicit Web UI allowlist remains restrictive. The broader catalog is used only when no allowlist is configured.
- Custom-provider readiness requires an enabled provider with an actual stored key.

## Automated verification

- Focused core/Web/Admin/custom-provider/security tests: **63 passed, 0 failed**.
- Full root suite passed once: **3,818 tests, 0 failed, 135 skipped**.
- A post-review repeat hit one unrelated OpenCode zero-second startup-output timing assertion; `test/opencode-harness.test.ts` passed immediately in isolation: **10/10**. No OpenCode code was changed.
- `npm run format:check`
- `npm run typecheck`
- `npm run lint`
- `npm run lint:ox`
- CLI package build + `npm run typecheck:contract`
- Deployment contract test: **1 passed**.
- Independent fresh-context review: no blocking findings. Its one P2 finding (stale readiness after custom-provider mutations) was fixed and covered before this commit.

## Live production-style verification

I reloaded this branch with the repository's dev supervisor and tested through Chrome against a saved DeepSeek OpenAI-compatible provider.

- Portal `/` returned the Web app instead of redirecting to onboarding.
- Admin reported `Ready`, showed `deepseek-v4-flash · custom provider key`, and listed the stored key only as `set (write-only)`.
- The picker displayed `DeepSeek V4 Flash` under `Pi` without an explicit Web model allowlist.
- Text turn returned exactly `DS_PR_TEXT_OK_20260810`.
- Tool turn called `execute` with `pwd`; Admin recorded exit `0`, the next DeepSeek request contained tool result `/root/workspace`, and the model returned `DS_PR_TOOL_OK`.
- Evidence session: `4de03b0b-6ec4-426d-aa72-8fd066cf9246`.

Admin reports the custom provider as ready

DeepSeek completes text and tool turns in the Web UI

## Contribution note

I understand `CONTRIBUTING.md` normally asks bug reporters to open issues rather than implementation PRs. Both bugs already have focused issues, and this patch is intentionally narrow, independently reviewed, and implementation-ready so maintainers can review, merge, or cherry-pick it directly. I am happy to adapt it to the maintainers' preferred shape.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/312"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788946268&installation_model_id=19911&pr_number=312&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F312&signature=e48ac4582f849679a38a27c9a1f94b341cee78767bc20edb7c2a7f1b6ac10a9b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


time-attack:
PR #298 is the narrow production wiring fix. If #298 lands, please rebase this PR and remove the duplicate entrypoint hunk while retaining the distinct protocol-safe picker and end-to-end custom-provider work. @16francej, can you confirm that split?

time-attack (commented):
I reproduced #236 on current `origin/main` (`76f8632`), applied this commit cleanly on top, and ran the focused core, Web, and Admin suites plus root/Web/Admin typechecks, ESLint, Oxlint, and Prettier. The protocol metadata boundary is directionally right and does not expose custom provider keys or base URLs, but I do not think the current broad patch is merge-ready:

1. **P1 — custom-provider refresh is process-local.** `src/index.ts` wires `refreshCustomProviders`, but that callback only updates the current process's `setCustomProviders` registry. An Admin mutation refreshes the replica that handled the request; other blue-green/multi-instance replicas continue using the old provider catalog indefinitely, so picker contents and turns become replica-dependent. Initial hydration is also fire-and-forget before the server starts. Resolve/catalog custom providers through the durable store or add a durable cross-process version/refresh mechanism, and complete initial hydration before accepting traffic.

2. **P2 — readiness tests ciphertext presence, not a usable key.** `CustomProviderStore.statuses()` reports `hasKey: Boolean(apiKeyEnc)`, while the existing corrupt-key test proves `resolveKey()` can fail with `hasKey === true`. The added Portal/Admin readiness checks can therefore report Ready and bypass onboarding even though turns cannot obtain credentials. Derive a non-secret usable/unreadable status at the store boundary and consume it consistently.

3. **P2 — zero token limits are accepted but not preserved.** Validation permits `contextWindow: 0` and `maxTokens: 0`, while both `src/api/routes/surface.ts` and `plugins/web-ui/src/pi-models.ts` use truthiness when copying them. Either require positive integers or copy defined values with `!== undefined`.

4. **Coverage/QA gap — only the OpenAI-compatible path is exercised in the browser evidence.** Add an Anthropic-compatible case through runtime option construction and transcript/tool activity reconstruction, asserting provider/api identity across the tool-result continuation. Rerun the required live QA in Firefox; the checked-in evidence explicitly says Chrome.

Scope-wise, #303 is the materially smaller #236 picker fix and #298 is the narrow production-wiring fix. #333 should not replace this metadata design because it sends provider base URLs to the browser. My recommendation is to split this PR along those boundaries, rebase after the narrow wiring work, and keep any remaining readiness/default-allowlist changes independently reviewable.

ReganBell:
Thanks for this. The end-to-end wiring landed on main via #512 (server-deps assembly, lazy boot default, picker support for custom slugs), which covers what this PR addressed. Closing as superseded.

PR #336, opened 2026-08-11
feat(slack): support native agent streaming

MiscMich:
## Summary

- publish agent thinking status and reply text through Slack's native agent APIs
- render durable QM tasks as native timeline updates without exposing reasoning or tool arguments
- preserve the existing reaction, Block Kit, and final-message delivery path when native streaming is unavailable
- delete an uncheckpointed stream before falling back so recovery cannot duplicate a reply
- let the seeded GitHub/GitLab skill use a per-user `VAULT_TOKEN_API_GITHUB_COM` as command-scoped `GH_TOKEN`, without printing or persisting it

## Verification

- `npm run typecheck`
- `npm run lint`
- `npm run lint:ox`
- `npm run format:check`
- focused Slack presenter, integration, delivery, messaging, turn-stream, and seeded-skill tests
- latest seeded-skill suite: 11/11 passed

## Live proof

- Slack persisted the native agent reply with `streaming_state: completed`
- the isolated sandbox executed `command -v gh` and `gh --version` and returned `/usr/bin/gh` / `gh 2.97.0`
- no GitHub connector token was present, proving that no host credential was silently inherited
- the final Slack UI rendered the response as an `AGENT` reply in the DM thread
- the prompt explicitly prohibited authentication and external mutations

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/336"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/skills-seed/github-gitlab/SKILL.md
+++ b/skills-seed/github-gitlab/SKILL.md
@@ -12,9 +12,20 @@ requiredCapabilities:
 Use this skill when the user asks to inspect repos, issues, pull requests, merge
 requests, code history, branches, or to make a small code change in a hosted repo.
 
-This is a resident-machine-auth connector. Prefer the native CLIs (`gh`, `glab`) and
-`git`, using the agent computer's logged-in state. Do not ask the user to paste tokens,
-and do not rely on proxy bearer-token injection.
+Prefer the native CLIs (`gh`, `glab`) and `git`. GitHub supports either the requesting
+user's product OAuth token or resident machine auth. When
+`$VAULT_TOKEN_API_GITHUB_COM` is present, pass it to `gh` only through `GH_TOKEN` for
+the command being run:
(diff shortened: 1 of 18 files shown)


time-attack:
Thanks for this, @MiscMich — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/737 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

Issue #350, opened 2026-08-12
Tools that install a program don't work on AWS — the agent reports the command doesn't exist


shekharkadyan:
## What happens

You add a tool to your QM deployment. It installs a command-line program — say
your company's search tool — using the `sandbox/Dockerfile` that QM's own
documentation tells you to write.

`qm check` says everything is valid. The deploy succeeds. Then someone asks the
agent to use it, and the agent replies that the command doesn't exist.

Nothing warned you. Nothing failed. The tool simply isn't there.

## Why

Your tool's program has to physically be inside the sandbox. There are two ways
QM builds sandboxes, and only one of them uses your Dockerfile:

- **On Fly (`sprites`)**, `qm sandbox publish` builds a sandbox image *from your
  Dockerfile*. Your program gets installed. This works.
- **On AWS**, the sandbox is built from QM's own template. Your Dockerfile is
  never opened.

What QM does send to AWS is a description of your tool — its name, its rules,
its help text — but not the program itself. So QM tells the agent "you have a
command called X", and X was never installed.

`qm check` passes because it only asks *"is there a Dockerfile?"* — not *"will
this Dockerfile ever be built?"* On AWS the answer to the second question is no.

## It isn't just Dockerfile-installed tools

The other documented way to give a tool its program — committing an executable
file next to `tool.json` — **also doesn't reach an AWS sandbox.**

What QM uploads for each tool is exactly one file: `tool.json`. Sibling files in
the tool's folder are never included, and the upload only accepts text, so a
compiled program couldn't travel that way even if it were.

That executable path does work on Fly, because `qm sandbox build` copies the
folder into the sandbox image. On AWS there is no image build to copy into.

So on AWS, **no tool that needs its own program can work** — regardless of how
you package it.

## Why this matters

Anyone building a tool on Fly and later moving to AWS will find it stops
working, with no error pointing at the cause. The agent just says the command
is missing, which looks like the tool author's fault.

## How to see it yourself

You don't need a full deployment:

1. `qm init --target aws`, and add any tool that relies on a Dockerfile.
2. `qm check` — passes.
3. `qm infra build-image`, then look at the file it uploads to S3.

That upload contains exactly two files: QM's own `Dockerfile` and its
`agent.mjs`. Nothing from your `sandbox/` folder is in it.

## What would fix it

Any one of these:

- Build your `sandbox/Dockerfile` into the AWS sandbox image, the way Fly does.
- Give tools a real way to declare an install step that works on AWS. (Today
[...]

Issue #272, opened 2026-08-07
sandbox.image is silently ignored when backend is "sprites"


bassemammar:
With `sandbox.backend: "sprites"`, a configured `sandbox.image` is accepted by `qm check` and then never used. `src/sandbox/sprites-sandbox.ts` has no image or rootfs code path at all — only `/v1/sprites/{name}/exec` and `/v1/sprites/{name}/policy/network`. The only consumers of a configured image are `local-sandbox.ts` and the `aws` branch of `config.ts`.

The sprite boots the stock base instead, which is hard to spot because it presents as success. `uname -a` returns `Linux sprite 6.12.91-fly ... x86_64` with exit 0, so the box is clearly alive. What gave it away was `df` in the same output: 2.4 MB used on the overlay, against an image that is 3.71 GB. The API corroborated it — `version` and `environment_version` were both null on every sprite.

Erroring at config load ("sandbox.image is not supported with backend: sprites") would turn a silent misconfiguration into an obvious one.

Version: 0.1.4

fo82:
We see this issue too. Workaround is to have a skill that checks the state of the sprite and installs/updates resources you need on there.

Issue #354, opened 2026-08-12
Correcting a placeholder accountId after apply makes Terraform replace the bucket holding all agent files


shekharkadyan:
> **Correction (see comments):** the trigger is not "any config change". The
> bucket name is derived from `accountId`, `region` and `cluster` only. What I
> actually hit was applying with the scaffolded placeholder `accountId`
> (`000000000000`) still in place, then correcting it. The body below is left
> as filed; read the comment for the accurate version.

## What happens

You make a routine change to your QM config — in my case, removing `slack` from
the list of services. You re-render and run `terraform plan`.

The plan wants to **delete and recreate your S3 bucket**.

That bucket holds every agent's saved files. If you don't notice, you lose all
of them.

## Why

The bucket's name is derived from a hash of your configuration. Change the
config, and the name changes:

```
before : qm-pipeshub-aws-b98cb4663ec1
after  : qm-pipeshub-aws-ae9e7d3e3353
```

Terraform can't rename a bucket, so its only option is to destroy the old one
and create a new one. It plans exactly that:

```
aws_s3_bucket.objects                          delete, create
aws_s3_bucket_lifecycle_configuration.objects  delete, create
aws_s3_bucket_policy.objects                   delete, create
aws_s3_bucket_versioning.objects               delete, create
```

## Why this matters

Nothing in that output says "this is where agent files live." It reads like
ordinary churn — the kind of thing you skim past in a plan you've already read
twice. And the change that triggered it had nothing to do with storage.

To catch it you'd have to both read the plan closely *and* already know what
that bucket contains.

There is a fix once you know: setting `aws.objectStoreBucket` pins the name. In
my case that took the plan from 7 destroys down to 2. But it's an optional
setting most people won't know exists until after it's too late.

## What would fix it

Any of:

- Derive the bucket name from something stable — the org or cluster name —
  rather than a hash of the whole config.
- Have `qm infra render` refuse to change the bucket name once a deployment
  exists, and tell the operator to pin `aws.objectStoreBucket`.
- Set `prevent_destroy` on the bucket in the reference module, so this fails
  loudly instead of proceeding quietly.

The last one is a one-line change and would turn silent data loss into an error
message.

time-attack:
I cannot reproduce this, can you please send me your version and any logs ?

shekharkadyan:
You're right that it doesn't reproduce as I described it, and I got the cause wrong. Sorry — here's the correct version.

[...]

PR #356, opened 2026-08-12
adrs: agents are much more useful with a company context layer

shekharkadyan:
Adds `adrs/agent-context-layer.md`, per CONTRIBUTING.

Two things in it:

1. A view we'd like yours on — whether "context layer" is a pattern QM wants to support as a documented thing deployments can have.
2. Two bugs blocking our implementation of it, both already filed: #272 and #350.

Happy to test against a branch.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/356"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789116415&installation_model_id=19911&pr_number=356&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F356&signature=76de60696cbe01f1f57024817e1acae93a67d7705e8b160d9f6fe6726429e522"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/adrs/agent-context-layer.md
+++ b/adrs/agent-context-layer.md
@@ -21,6 +21,20 @@ a PipesHub thing.
 attachment. Each person supplies their own token from their own keychain, so
 answers stay bounded by their own access.
 
+**To be precise about that last part:** it holds in DMs, and we've only built
+for DMs. In a personal scope QM injects the speaker's own keychain credential,
+so the person asking is the person whose permissions apply — we've confirmed
+that end to end. In a shared conversation it works differently: your own
+credential isn't injected, someone grants one to the conversation instead, and
+that grant then serves whoever speaks next. For us that would mean one person's
+token answering another person's question, which is the exact thing this design
+exists to prevent. So we've scoped v1 to personal DMs and say so in our docs.
+
+Supporting shared rooms properly would need QM to broker per speaker — inject
+the credential belonging to whoever is talking, rather than one attached to the
+room. We're not asking for that today; flagging it as the thing that would
+unlock it.
+
 **What we're asking:**
 
 1. **Fix the two bugs below** — they're the only thing stopping this working.
--- a/adrs/agent-context-layer.md
+++ b/adrs/agent-context-layer.md
@@ -1,0 +1,51 @@
+# Agents are much more useful with a company context layer — and two bugs are blocking it
+
+Hi — we're PipesHub. We do context layer for agents and enterprise search for humans across Drive, Slack, Gmail, Jira,
+Confluence and internal knowledge bases, with each entities's permissions enforced
+on every query.
+
+**Why we think this matters for QM.** A QM agent with a sandbox can already do
+real work. What it can't do is know anything about the company it works for — so
+people spend the first half of every conversation pasting in context, and the
+agent still reasons from a fragment. The obvious fix is to connect data sources.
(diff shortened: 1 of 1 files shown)


adrs/agent-context-layer.md
```diff
@@ -0,0 +1,51 @@
...
+than one that reads exactly what the person asking can already read.
+
+A context layer solves the second problem, which is what makes solving the first
+one safe. We think that's a general pattern worth QM having a view on, not just
+a PipesHub thing.
+
+**What we built.** A CLI installed into the sandbox, plus a `tool.json` and a
+`SKILL.md`. Deliberately QM-native — no core changes, no plugin, no MCP
+attachment. Each person supplies their own token from their own keychain, so
+answers stay bounded by their own access.
```
time-attack, line 22: QM only injects a person’s own keychain automatically in their personal scope. In an ordinary shared conversation, the speaker’s own credential is not injected; credentials must instead be granted to the conversation, and a standing grant can then be reused by subsequent participants. I reproduced this with the focused keychain end-to-end tests and confirmed the controlling condition in `src/core/orchestrator.ts`. As written, “answers stay bounded by their own access” is therefore true for DMs but not established for shared rooms, which is important because permission preservation is the proposal’s central safety claim. Could this describe the per-speaker credential-brokering mechanism that prevents cross-participant reuse, or explicitly limit the integration to personal DMs?

shekharkadyan:
@time-attack Good catch, and you're right. I've pushed a correction.

I confirmed it in `src/core/orchestrator.ts` — `materializeOwn()` only runs when `scopeId === personalScope(actor.id)`, and everything else comes from `materializeStanding(scopeId)`, which is attached to the conversation rather than the speaker.

So the sentence was true for DMs and unsupported for shared rooms, and you're right that it's the wrong sentence to be loose about — permission preservation is the whole point of the design.

Taking your second option: **explicitly limited to personal DMs.** That's what we'd already scoped v1 to and documented on our side; the note just didn't say it.

The reason we haven't tried to work around it: in a shared room a standing grant would mean one person's token answering another person's question, which returns documents the asker can't see. That's precisely the failure the per-person model exists to prevent, so a partial version seemed worse than none.

I've noted that shared-room support would need brokering per speaker — injecting the credential belonging to whoever is talking, rather than one attached to the room. Not asking for it today, just flagging it as the thing that would unlock it.

Separately, thank you for reproducing it rather than taking the claim at face value. On #354 I'd got the cause wrong in the same way and your colleague's "I cannot reproduce this" prompted me to re-test — the trigger turned out to be much narrower than I'd written. Correction posted there.

PR #362, opened 2026-08-12
[codex] Preserve auth email allowlist with domain

### LINKED ISSUE #328: Setting AUTH_ALLOWED_EMAIL_DOMAIN silently drops AUTH_ALLOWED_EMAILS, locking out every address outside the domain
Setting `AUTH_ALLOWED_EMAIL_DOMAIN` makes the CLI stop shipping
`AUTH_ALLOWED_EMAILS` entirely, so every address outside the domain loses
sign-in.

How I ran into it, on the docker target: I added the domain so colleagues at our
own domain wouldn't need a redeploy each time, kept the existing address list in
`.env`, and ran `check` then `up`. Both passed. The list was simply gone from the
auth container afterwards — and my own address is a gmail one, so the next
sign-in would have been mine, failing. I only noticed because I inspected the
container's env.

The auth broker ORs the two (`plugins/auth/src/server.ts`), so both are meant to
coexist; the CLI treats them as mutually exclusive:

```js
// secrets.js
function requirementFor(config, spec) {
  if (conditionMatches(config, spec.required.when)) return true;
  return spec.required.optionalOtherwise ? false : null;
}
// computedSecrets: if (required === null) continue;
```

`AUTH_ALLOWED_EMAILS` is declared `required: { when: env-absent
AUTH_ALLOWED_EMAIL_DOMAIN }` with no `optionalOtherwise`, so once the domain is
set `requirementFor` returns null and the secret is dropped before
`computedSecrets` ever returns it — not made optional, removed. `fly.js` stages
from that same list, so it is never `fly secrets set` on the app.

On fly this splits by deploy order rather than failing outright:

- domain added to an **existing** app — the old `AUTH_ALLOWED_EMAILS` secret is
  still on the app, since nothing unsets it (only `SECURITY_SCREEN_PROXY_TOKEN`,
  `FLY_DEPLOY_API_TOKEN` and `FLY_BASE_IMAGE` get an explicit unset), so it keeps
  working
- **fresh** app from the same config — never staged, so only the domain applies
  and outside addresses are locked out

Same repo, same `.env`, different behaviour, and nothing reports it: `qm check`
passes, `qm up` succeeds, and the staging output lists what was staged, not what
was dropped. The failure lands on whoever is outside the domain — often the
operator's own personal address.

The portal copy has it too: `AUTH_ALLOWED_EMAILS` also aliases to
`OIDC_ALLOWED_EMAILS` for the portal, gated on the same condition, so an OIDC
deployment loses its out-of-domain addresses the same way.

Workaround, if it helps anyone: put both under `env.auth` in `qm.config.jsonc`
so they bypass the secret schema. On fly, a stale secret of the same name will
still win over `[env]`.

## Suggested fix

Give both `AUTH_ALLOWED_EMAILS` specs `optionalOtherwise: true`, mirroring
`ANTHROPIC_API_KEY` and `OPENROUTER_API_KEY` (`cli/src/secrets.ts:45,52`):

```js
{
  name: "AUTH_ALLOWED_EMAILS",
  service: "auth",
  required: {
    when: { kind: "env-absent", service: "auth", name: "AUTH_ALLOWED_EMAIL_DOMAIN" },
    optionalOtherwise: true,
  },
}
```

With that, `requirementFor` returns `false` (optional) instead of `null`
(dropped) when the domain is set, so a list the operator kept in `.env` is still
staged and shipped — which is what the broker's OR already expects. The same
change on the portal spec keeps `OIDC_ALLOWED_EMAILS` in step.

A regression test would set `AUTH_ALLOWED_EMAIL_DOMAIN` alongside a provided
`AUTH_ALLOWED_EMAILS` and assert the list still appears in `computedSecrets`
and reaches both the auth and portal services.

time-attack:
Fixes #328.

## What changed

- keep `AUTH_ALLOWED_EMAILS` optional when `AUTH_ALLOWED_EMAIL_DOMAIN` is configured, so the CLI still sends it to auth and portal
- make portal authorization accept an address matching either the explicit list or the configured domain, matching the auth broker
- cover optional secret routing and combined list/domain authorization with focused regressions

## Root cause

The conditional secret specs had no `optionalOtherwise`, so a configured domain removed `AUTH_ALLOWED_EMAILS` from `computedSecrets` instead of making it optional. The portal also evaluated simultaneous list and domain rules as an intersection while the broker evaluated them as a union.

## Validation

- `node --test test/auth-broker.test.ts` in `cli` (8 passed)
- `node --test test/oidc.test.ts` in `plugins/portal` (14 passed)
- `npm run typecheck`
- `npm run lint`
- Prettier check on changed files
- independent fresh-context review: no findings

Live dev-instance QA was attempted, but no Slack pool app became available during the launcher's bounded wait, so no live instance was created.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/362"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789171146&installation_model_id=19911&pr_number=362&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F362&signature=2a652fcdfadb9f9350b60d29ad491b48bde114f2ee18fb118e44461f7c4fe3b4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yc-software/qm",
-  "version": "0.1.6",
+  "version": "0.1.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@yc-software/qm",
-      "version": "0.1.6",
+      "version": "0.1.7",
       "license": "MIT",
       "bin": {
         "qm": "dist/bin/qm.js"
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yc-software/qm",
-  "version": "0.1.6",
+  "version": "0.1.7",
   "license": "MIT",
   "description": "Control-plane CLI for portable QM deployments on Docker, Fly, and AWS.",
   "type": "module",
--- a/cli/src/secrets.ts
+++ b/cli/src/secrets.ts
@@ -316,7 +316,10 @@ export const FIRST_PARTY_SECRET_SPECS: readonly SecretSpec[] = [
   {
     name: "AUTH_ALLOWED_EMAILS",
     service: "auth",
-    required: { when: { kind: "env-absent", service: "auth", name: "AUTH_ALLOWED_EMAIL_DOMAIN" } },
+    required: {
+      when: { kind: "env-absent", service: "auth", name: "AUTH_ALLOWED_EMAIL_DOMAIN" },
+      optionalOtherwise: true,
+    },
     description: "Comma-separated email addresses allowed to sign in through the built-in broker.",
   },
   {
@@ -331,6 +334,7 @@ export const FIRST_PARTY_SECRET_SPECS: readonly SecretSpec[] = [
           { kind: "env-absent", service: "auth", name: "AUTH_ALLOWED_EMAIL_DOMAIN" },
         ],
       },
+      optionalOtherwise: true,
     },
     description: "Email addresses allowed to sign in; the portal enforces the same list the broker does.",
   },
--- a/cli/test/auth-broker.test.ts
+++ b/cli/test/auth-broker.test.ts
@@ -137,7 +137,10 @@ test("the broker's generated secrets reach both sides under the right names", ()
   assert.ok(!names.has("OIDC_CLIENT_SECRET"), "the operator is never asked for an OIDC client secret in broker mode");
   assert.ok(!names.has("OIDC_CLIENT_ID"));
   assert.ok(!names.has("PORTAL_EXPECTED_TEAM_ID"));
-  assert.ok(!names.has("AUTH_ALLOWED_EMAILS"), "a configured domain removes the per-address allowlist requirement");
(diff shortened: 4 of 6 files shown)


time-attack:
The branch is mergeable, but GitHub currently shows only a skipped check. Before merging, please run the current CI matrix, complete live combined-domain/list authentication QA, and obtain the required fresh-context review. @16francej, are there any additional auth-policy cases you want covered?

PR #374, opened 2026-08-13
Allow certified read-only turns under strict posture

hudsonshank:
## Problem

Under strict posture every harness tool call pauses for human approval. That is correct for effectful work, but an explicitly read-only turn has already had its effectful tools removed before the harness starts. The approval gate then pauses tools that cannot do anything, which makes read-only strict deployments impractical without weakening posture.

## Change

Adds an opt-in read-only certification to the harness adapter contract and lets the orchestrator skip the approval gate only when the turn is already read-only.

- `HarnessAdapterProfile` gains an optional `readOnlyToolProfile?: true`. It is a certification an adapter makes about its own tool surface, not a runtime toggle, and the type admits only `true` so it cannot be set to a falsy placeholder.
- `requiresToolApproval(policy, enforcedReadOnly)` in `src/security/security-posture.ts` centralizes the rule: approval is required when `toolApprovals === "all"` and the turn is not an enforced read-only turn.
- The orchestrator installs `toolApprovalGate` unless the turn is strict-read-only **and** the selected harness certifies the read-only profile. Both conditions must hold.
- `createHarnessRouter` composes conservatively: the router certifies only when *every* selectable adapter certifies, and otherwise clears the flag rather than inheriting it from the utility profile.
- The strict posture prompt now states that an explicitly read-only turn does not pause, so the model's description of its own constraints stays accurate.

All in-tree adapters (claude, codex, opencode, pi, mock) declare the certification.

## Posture is preserved

Strict posture continues to require approval for every non-read-only tool call. The gate is omitted only where there is nothing effectful left to gate. Hard denials, authentication, authorization, tenant boundaries, credential scope, revocation, and audit are untouched, as is every non-strict posture.

The failure direction is deliberate: an adapter that says nothing gets the gate, and a router with one uncertified adapter gets the gate.

## Tests

- `test/security-posture.test.ts` covers `requiresToolApproval` across both postures and both read-only states.
- `test/harness-adapter.test.ts` covers router composition, including the mixed case where one adapter lacks the certification.
- `test/orchestrator.test.ts` covers gate installation and omission.

134 tests pass across the three affected files, plus `tsc --noEmit` and ESLint on the touched paths.

No screenshot: the change has no rendered surface.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/374"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789223368&installation_model_id=19911&pr_number=374&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F374&signature=54270bb4e4464d24c6057d97a61b18c9117231d73838c249accebc0ae04d13cb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -54,6 +54,7 @@ import type { GapWork, HarnessLlmRequestRecord } from "../harness/harness.ts";
 import { forModelContext } from "../harness/context-compaction.ts";
 import {
   renderSecurityPolicyPrompt,
+  requiresToolApproval,
   securityScreenPayload,
   UNSCREENED_REASON,
   unscreenedNotice,
@@ -2145,7 +2146,12 @@ export function createOrchestrator(deps: OrchestratorDeps): Orchestrator {
                   },
                 }
               : {}),
-            ...(securityPolicy.toolApprovals === "all" ? { toolApprovalGate: authorizeToolCall } : {}),
+            ...(requiresToolApproval(
+              securityPolicy,
+              strictReadOnly && deps.harness.profile.readOnlyToolProfile === true,
+            )
+              ? { toolApprovalGate: authorizeToolCall }
+              : {}),
             systemPrompt,
             systemCacheBoundary: stableSystemBytes,
             history: continuation?.history ?? history,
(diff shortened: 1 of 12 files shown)


hudsonshank:
Closing - opened in error, not intended for upstream contribution.

PR #379, opened 2026-08-13
Reject custom tools on AWS MicroVM sandboxes

### LINKED ISSUE #350: Tools that install a program don't work on AWS — the agent reports the command doesn't exist

time-attack:
## Summary

- reject custom tools and `sandbox/Dockerfile` when an AWS deployment uses Lambda MicroVM sandboxes
- preserve custom sandbox image support for AWS deployments using the sprites backend
- make AWS initialization scaffold a valid skill-only sandbox and document the backend limitation
- validate the sandbox before `infra build-image` can make AWS calls
- bump `@yc-software/qm` to 0.1.7

Fixes #350.

## Verification

- `node --experimental-test-module-mocks --test cli/test/check.test.ts cli/test/sandbox-build.test.ts cli/test/init.test.ts cli/test/cli-dispatch.test.ts cli/test/infra-image.test.ts`
- `node --experimental-test-module-mocks --test cli/test/package.test.ts test/release-workflows.test.ts`
- `npm --prefix cli run typecheck`
- `npm run lint`
- Prettier check for all changed files
- independent fresh-context review with no remaining findings

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/379"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789234537&installation_model_id=19911&pr_number=379&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F379&signature=2ce384a1d1a897265147e4cabb8b87a3dff43041c72bb49848781df90ce1e5a0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -79,6 +79,12 @@ building anything, an existing deployment manifest and no `sandbox.image` overri
 that override only seeds the first `qm up` and must be removed afterwards. Every
 ordinary `up` also syncs the layer.
 
+AWS Lambda MicroVM sandboxes use QM's fixed runtime image. They accept skills through the
+deployment layer, but do not build `sandbox/Dockerfile` or copy tool executables. `check`,
+`sandbox build`, and `infra build-image` reject custom tools and Dockerfiles for that
+backend. Configure `sandbox.backend: "sprites"` when the deployment needs a custom sandbox
+image.
+
 Auto uses its built-in model classifier unless `qm.config.jsonc` declares one
 `securityScreen` proxy with a provider label, HTTPS endpoint, and `shadow` or
 `enforce` rollout. The proxy token is routed separately through
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yc-software/qm",
-  "version": "0.1.6",
+  "version": "0.1.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@yc-software/qm",
-      "version": "0.1.6",
+      "version": "0.1.7",
       "license": "MIT",
       "bin": {
         "qm": "dist/bin/qm.js"
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yc-software/qm",
-  "version": "0.1.6",
+  "version": "0.1.7",
   "license": "MIT",
   "description": "Control-plane CLI for portable QM deployments on Docker, Fly, and AWS.",
   "type": "module",
--- a/cli/src/backends/registry.ts
+++ b/cli/src/backends/registry.ts
@@ -296,7 +296,7 @@ const aws: HostingProvider = {
   publishSandbox: async (ctx, opts) => {
     if (ctx.config.sandbox?.backend !== "sprites") {
       throw new CliError(
-        `this AWS deployment runs Lambda MicroVM sandboxes (sandbox.backend is not "sprites"); use \`qm sandbox build\` to validate the layer and \`qm infra build-image\` to publish the runtime — or set "sandbox.backend": "sprites" with "sandbox.app" to host sandboxes in an operator-published layer image`,
+        `this AWS deployment runs Lambda MicroVM sandboxes (sandbox.backend is not "sprites"), which do not support custom sandbox images; set "sandbox.backend": "sprites" with "sandbox.app" to host sandboxes in an operator-published layer image`,
       );
     }
     if (!opts.dryRun) assertAwsSandboxPinRecordable(ctx.config);
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -455,6 +455,7 @@ async function dispatch(argv: string[]): Promise<void> {
         if (operation === "delete-image") await deleteAwsMicrovmImage(ctx.config);
         else await deleteAwsTaskDefinitions(ctx.config);
       } else if (operation === "build-image") {
+        runChecks(ctx.config, ctx.configDir, ctx.sandboxDir, { report: false });
         await buildAwsMicrovmImage(ctx.config, ctx.configPath);
       } else {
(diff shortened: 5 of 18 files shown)


time-attack:
Coordination hold: #427 rejects the AWS-to-Sprites custom-image path because the installed Sprites SDK cannot materialize it. This PR must not merge as written. Its AWS MicroVM tool behavior should be revised to fail closed without recommending Sprites, then rebased; #427 does not close this PR.

shekharkadyan:
We filed #350, so thank you for picking it up. Making AWS say no is much better than the current behaviour, where the deploy succeeds and the agent just reports that the command does not exist.

Two notes from the downstream side, since we are probably the concrete case this is designed against.

## A scaffold that fails its own validation

We ship an installable integration that people scaffold into their QM deployment. It writes exactly the two things this PR rejects on MicroVM:

```
sandbox/tools/pipeshub/tool.json
sandbox/Dockerfile
```

Today, someone on AWS who runs our scaffold gets a deployment that comes up and an agent that cannot find the CLI — bad, and the reason #350 exists. After this lands they get a hard validation failure instead, which is better, but the message points at the files rather than at what to do next.

The person in that position has not done anything wrong: they followed our guide, and our guide already tells them to use `sprites`. It would help if the error named the way out — something like "AWS MicroVM sandboxes run skills only; move the tool's guidance into a SKILL.md, or use the sprites backend." Right now "remove it" reads as "your integration does not work here", when the accurate version is "the binary cannot be installed here, and the rest of it still can."

We will handle our side either way — the scaffold can stop writing those files when the target is AWS. Mostly flagging that "reject" plus a scaffold that emits the rejected files is a combination worth catching before release.

## The same concern as on #427

This rejects custom **tools**, and a tool descriptor is where `egress`, `auth`, and `approvals` live (`sandbox-layer.d.ts:26-33`). Those need nothing installed — they are declarations the platform enforces. There is no skill-side equivalent for them.

For us that means an AWS deployment loses the network allowlist and the approval rules that make this safe to hand to another organization, not just the binary we already install from the skill's first-run step.

Same suggestion as there: reject `install` and the Dockerfile, keep a descriptor that carries only `egress` / `auth` / `approvals` / `hints`. Happy to help test whichever shape you pick.

Issue #225, opened 2026-08-05
Web UI composer: Enter that confirms IME composition sends the message (Japanese/Chinese/Korean input)


toyoshi:
## What happened

Typing Japanese in the web UI composer, pressing Enter to confirm the IME
(input method editor) conversion immediately sends the message, with the
composition text still half-finished. This affects every composition-based
input method — Japanese, Chinese (pinyin etc.) and Korean at least — so the
composer is nearly unusable for CJK users.

## What I expected

Enter during IME composition should only confirm the conversion; only a
subsequent Enter should send the message.

## Steps to reproduce

1. Open the web UI composer in Chrome on macOS with a Japanese IME enabled.
2. Type romaji, e.g. `konnichiha` — the composer shows the underlined
   composition text.
3. Press Enter to confirm the kanji/kana conversion.
4. The message is sent immediately instead of just committing the text.

## Additional context

The keydown handler in `plugins/web-ui/src/composer.ts` treats every
non-shift Enter as "send" without checking `e.isComposing` (or the legacy
`keyCode === 229`).

This is a well-known pitfall that other TypeScript projects guard against:

- **Element Web** (Matrix messenger): the composer has an `isComposing()`
  helper checking `event.nativeEvent.isComposing` (plus composition events and
  a Safari-specific `event.which == 229` fallback), and both the editing and
  send paths return early while composing:
  https://github.com/element-hq/element-web/blob/develop/apps/web/src/components/views/rooms/BasicMessageComposer.tsx
  https://github.com/element-hq/element-web/blob/develop/apps/web/src/components/views/rooms/SendMessageComposer.tsx
- **VS Code**: normalizes this at the framework level — `keyboardEvent.ts`
  maps `e.isComposing` to `KeyCode.KEY_IN_COMPOSITION` (keyCode 229) so no
  keybinding ever fires mid-composition:
  https://github.com/microsoft/vscode/blob/main/src/vs/base/browser/keyboardEvent.ts

Environment: qm 0.1.4 (Fly deployment), Chrome on macOS, Japanese IME.

PR #89, opened 2026-08-01
Mid-run steer drops attachments

mariohercules:
Adding an ADR in `adrs/`, per CONTRIBUTING. Same thing I reported in #48, written up here as
text rather than as a bug ticket.

Short version: a message sent while a run is live only steers its text — attachments are
dropped, and an attachment-only message is treated as empty and disappears entirely, with the
Slack handler standing down because the turn still reports `steered: true`. Easy to hit,
since it's the shape the agent asks for when it says "send me the screenshot".

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/89"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788177409&installation_model_id=19911&pr_number=89&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F89&signature=82cf5b5f2bd8448c9256ba75618a98a3312c9045db28b2f989152dc400ecea86"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #391, opened 2026-08-13
Contributor-reported fixes: paths with spaces, redactor gaps, safe Slack

ReganBell:
cuts, jsonb-safe serialization, CAS promotion, mid-run files, IME Enter Seven first-principles fixes for reported bugs, one commit per bug. bench:memory and the twin-instance script broke under any directory containing a space (new URL(...).pathname keeps percent-escapes) — use fileURLToPath everywhere a file URL becomes a path (#77). The audit redactor missed base64url-shaped (JWT-alphabet) secrets and its --with-token rule replaced the match with itself, a literal no-op (#80). Slack section splitting, clip, and inlineCode cut with bare slice, bisecting emoji surrogate pairs, <url|label> entities, and format runs — one safe-boundary cutter now backs all three (#82). The Postgres-backed durable map JSON.stringifies into jsonb, which rejects NULs and lone surrogates the system's own truncation helpers manufacture — sanitize at the serialization boundary (#62). Scratch-memory promotion wrote back with a blind replace, reverting any edit that landed during its model call — the write is now compare-and-set against the revision read at the start (#79). Steering a live run with a captionless file dropped the file while reporting steered:true — wakes now carry attached file names so the file is announced to the run (#89). And the Enter that confirms an IME composition (Japanese/Chinese/Korean) no longer submits the half-composed message (#140, issue #225).  toyoshi

**Deployment notes**

Batch of small fixes; mostly behavior-neutral (fileURLToPath, base64url secret masking, Slack
safe-cut splitter, durable-map jsonb tweak).
Secret masking now redacts base64url tokens and --with-token payloads — audit lines get MORE
redaction; any org tooling that parsed raw tokens out of audit logs (unlikely, undesirable) loses
them.
Slack messages near the 2900-unit boundary split at slightly different points; cosmetic.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/391"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789237844&installation_model_id=19911&pr_number=391&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F391&signature=c27a6ff0974722bac9130ebc317938623848f6112d26d6c5c75bd474dc19e8b2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/web-ui/src/composer.ts
+++ b/plugins/web-ui/src/composer.ts
@@ -1078,6 +1078,10 @@ export function createComposerSurface(ctx: ConvCtx): ComposerSurface {
   }
 
   function onComposerKeydown(e: KeyboardEvent, agent: Agent): void {
+    // During IME composition (Japanese/Chinese/Korean), Enter confirms the
+    // conversion — it must never send. Safari reports composition Enter with
+    // keyCode 229 and may fire after compositionend, so check both.
+    if (e.isComposing || e.keyCode === 229) return;
     const slash = currentSlashMenu();
     if (slash.open) {
       if (e.key === "Escape") {
--- a/plugins/web-ui/src/sessions.ts
+++ b/plugins/web-ui/src/sessions.ts
@@ -971,6 +971,7 @@ function renameInput(menuKey: string, ariaLabel: string, commit: () => Promise<v
         renameDraft = (e.currentTarget as HTMLInputElement).value;
       }}
       @keydown=${(e: KeyboardEvent) => {
+        if (e.isComposing || e.keyCode === 229) return;
         if (e.key === "Enter") {
           e.preventDefault();
           void commit();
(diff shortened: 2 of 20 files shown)


time-attack:
CodeQL is currently failing and this PR combines several unrelated fixes. The simplest review path is to extract the #225 IME composition guard into a focused PR and keep the remaining changes in coherent units. If the combined scope stays, please resolve CodeQL and explain why the bundle should be reviewed atomically. @16francej, which direction do you prefer?

PR #414, opened 2026-08-13
Make tool call primary-key migration conditional and atomic

### LINKED ISSUE #46: Stop rebuilding the tool_calls primary key on every boot
## Problem

The run store's DDL list ends with an unconditional pair (`src/runs/postgres-run-store.ts:76-77`):

```sql
ALTER TABLE tool_calls DROP CONSTRAINT IF EXISTS tool_calls_pkey;
ALTER TABLE tool_calls ADD PRIMARY KEY (run_id, attempt, call_index);
```

These run on **every** startup rather than once. `tool_calls` has no retention — nothing deletes from it and `deleteSession` does not touch it — so every deploy takes an ACCESS EXCLUSIVE lock on an ever-growing table and rebuilds the index while the blue-green sibling is still serving `ledger.begin`/`record` against it.

There is also a correctness window. Between the DROP and the ADD there is no unique constraint, so the live instance's `record()` (`ON CONFLICT DO NOTHING`, `:333`) can commit a duplicate `(run_id, attempt, call_index)`. The next boot's `ADD PRIMARY KEY` then fails with a unique violation, and because schema init runs on startup, every subsequent boot fails until someone dedupes the table by hand.

The pair reads as a one-time migration for the added `attempt` column that was never made conditional. Fresh databases already get the correct primary key from the `CREATE TABLE` above it.

## Expected behavior

Schema initialization is idempotent and cheap on every boot. The primary key is corrected only on a database that actually has the old one, and never leaves the table without a unique constraint while another instance is writing.

## Acceptance criteria

- Booting against an up-to-date database issues no DDL that locks or rewrites `tool_calls`.
- A database with the pre-`attempt` primary key is still migrated exactly once.
- The migration cannot leave the table without a unique constraint while a sibling instance serves traffic.
- Tests cover both the already-migrated and the legacy-schema database.

time-attack:
## Summary

- replace the unconditional `tool_calls` primary-key rebuild with one catalog-gated transaction under the existing schema-init advisory lock
- leave current schemas untouched while migrating legacy, renamed-key, and interrupted no-key states without exposing a uniqueness gap
- cover current-schema writer concurrency, one-time legacy migration, interrupted-state recovery, and rollback when replacement-key creation fails

## Verification

- 9 affected PostgreSQL run-store tests
- TypeScript typecheck
- ESLint and Oxlint
- Prettier and `git diff --check`
- three independent fresh-context reviews with PostgreSQL, migration-test, and regression lenses

Fixes #46

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/414"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789243161&installation_model_id=19911&pr_number=414&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F414&signature=ead56013ba4470f62ef6841accc268600f508cef421bc75ea4cc41d0be29713f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


time-attack:
Closing — superseded. The one-shot transactional tool_calls PK migration landed on main in #494 (c58bbae).

PR #418, opened 2026-08-13
Fail closed on unsupported Fly application publishing

time-attack:
Closes #209.

## Summary

- reject unsupported `DEPLOY_PROVIDER` values during core configuration and CLI validation instead of silently selecting Docker
- preserve ordinary Fly hosting while routing an unset Fly application publisher to an explicit unavailable provider, so publish cannot spawn Docker
- remove the nonexistent Fly publisher token from runtime and CLI secret contracts while cleaning up previously staged copies
- fail cleanup closed when the original provider is unavailable so external runtimes are never falsely marked destroyed

## Verification

- 217 affected core, CLI, Fly backend, and Fly E2E tests
- 81 focused tests after rebasing onto the latest `main`
- root and CLI TypeScript checks
- ESLint and Oxlint
- Prettier and `git diff --check`
- two independent fresh-context reviews, with all findings resolved

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/418"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789243487&installation_model_id=19911&pr_number=418&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F418&signature=4c3b3c4fcb8862c2b9699314aeb7fd6a23d9c0ec1407256204241b038d9c98bb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #416, opened 2026-08-13
Enforce dependency cooldown during CLI bootstrap

time-attack:
Closes #42.

## Summary

- fail closed before registry resolution unless the installed npm supports `min-release-age`
- apply the seven-day release gate to every documented and generated `@latest` bootstrap command
- declare npm 11.10 as the CLI and generated deployment minimum and bump the CLI to 0.1.7
- cover an old npm preflight failure and releases immediately on both sides of the cooldown

## Verification

- bootstrap and init tests: 20 passed
- packed npm artifact tests: 2 passed
- deployment skill tests: 6 passed
- root and CLI TypeScript checks
- ESLint and Oxlint
- Prettier and `git diff --check`
- fresh-context independent review: no remaining findings

Live dev-instance, browser QA, and screenshots are not applicable because this changes CLI bootstrap commands and package metadata only, with no Slack or rendered UI behavior.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/416"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789243370&installation_model_id=19911&pr_number=416&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F416&signature=1815a331ca0006fba2b2eacc021c5884f26ebbfcd5805e3352e6bb5296cb0ce8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #427, opened 2026-08-13
Reject unsupported Sprites customization

### LINKED ISSUE #137: sprites backend: sandbox/tools binaries, secretEnv, and deployment layer never reach the sandbox (fly target) — silent failure
**Setup:** qm 0.1.4, target=fly, SANDBOX_BACKEND=sprites (the fly default), sandbox image published via `qm sandbox publish` (digest recorded in config; core env FLY_BASE_IMAGE matches), deployment layer reported synced (`deployment layer: v2 <hash>` on `qm up`).

**Symptom:** Sprites boot without any of the operator's sandbox customization:

- `sandbox/tools/*` binaries are absent — including the scaffold's own `example-tool`. `qm sandbox build` COPYs tools to `/usr/local/bin/`, but the live sprite's `/usr/local/bin` is empty, which suggests sprites are not booting the operator-published layer image at all (filesystem looks like the stock sprite base: `/.sprite/bin`, `/home/sprite`).
- `sandbox.secretEnv` values never appear in the sprite environment (no `FLY_RESIDENT_ENV_*`, no bare name).
- Agents report the advertised tools are not in their callable set either.

No errors are logged by core, `qm up`, or `qm doctor` — everything reports success; the content just never arrives. Verified by `sprite exec` into the live sandbox created moments earlier by core.

**Two adjacent paper cuts found while debugging:**
1. `qm secrets push` silently skipped `SPRITES_TOKEN` (required secret for the default backend) — it was never staged on core; `fly secrets set` directly was needed. A required-secret diff against the app would catch this.
2. `.env.example`'s `PUBLIC_API_URL` comment ("Public core self-API URL") reads as if a core hostname is expected, but core ships with no public service; the working value is the portal `publicUrl` (per `commands/setup.js` hints). Scripted setups (the documented `.env`-editing path) can easily get this wrong — a validation error or clearer comment would help.

**Expected:** either sprites boot the published layer image (so tools/secretEnv arrive), or `qm up`/`doctor` should fail loudly when the sprites backend cannot apply the recorded sandbox layer.

### LINKED ISSUE #272: sandbox.image is silently ignored when backend is "sprites"

time-attack:
Fixes #272
Fixes #137

## Summary

- reject Fly Sprites image, resident-environment, tool-binary, and direct `FLY_*` customization instead of accepting settings the installed SDK cannot materialize
- reject the removed AWS-to-Sprites image path while preserving Docker image customization and AWS MicroVM image management
- validate tool-bearing deployment layers at the durable store boundary whenever either active sandbox backend is Sprites, including persisted records and migration routing
- scaffold and document Fly as the stock Sprites runtime, while continuing to accept skill-only deployment layers

`SPRITES_TOKEN` discovery is intentionally unchanged; it is a separate credential contract from the silent customization paths fixed here.

## Coordination

- #379 must not merge as written: its suggested AWS MicroVM escape hatch switches to Sprites for a custom image, which this change rejects because the SDK has no materialization path. Its remaining AWS MicroVM tool question should be revised to fail closed without recommending Sprites, then rebased; this PR does not close #379.
- #416 overlaps only in claiming CLI version `0.1.7`. Whichever change lands second must rebase and take the next unused version.
- #418 is complementary: it removes unsupported Fly application publishing and `FLY_DEPLOY_API_TOKEN`, while this change removes the sandbox-image application path and `FLY_SANDBOX_API_TOKEN`. Whichever lands second must rebase and preserve both fail-closed contracts.

## Verification

- 212 CLI configuration, check, init, Fly, sandbox build/publish, doctor, and secret tests
- 49 runtime deployment-layer and Sprites profile tests
- packed npm artifact test across separate Fly, Docker, and AWS consumers
- root and CLI typechecks
- ESLint, Oxlint, Prettier, and `git diff --check`

No live browser QA or screenshots: this changes CLI/runtime configuration contracts and has no rendered Slack or web surface.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/427"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789244598&installation_model_id=19911&pr_number=427&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F427&signature=c1cfa3a293679e5742d788f31cbc2a0d3d5390ea043e2c1899d8f6bc9a5ecaa6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


shekharkadyan:
We filed #272, so first: thank you for taking it seriously. Turning a setting that was silently ignored into one that says no is the right instinct, and it is exactly what we asked for.

But as written this also removes three things that have nothing to do with images, and I do not think that is intended.

## What the change does beyond the image

`validateSandboxLayer` now rejects **every** tool descriptor when the target is `fly`, not just descriptors that need a binary:

```js
if (sprites) {
  out.errors.push(`tool "${descriptor.id}" cannot be installed in Fly Sprites ... remove it`);
}
```

The reason given is that the SDK cannot materialize tool binaries or a custom image. That is true, and rejecting `install` on that basis makes sense. The problem is that `install` is not all a tool descriptor carries. From `sandbox-layer.d.ts:26-33`:

```ts
export interface ToolDescriptor {
  install?: ...
  egress?: string[];
  auth?: ToolAuthDescriptor;
  approvals?: ToolApproval[];
}
```

`egress`, `auth`, and `approvals` do not need anything materialized. They are declarations the platform enforces at runtime. And they exist **only** on `ToolDescriptor` — there is no skill-side equivalent, so once tool descriptors are rejected, Sprites deployments lose them entirely.

## What that costs, concretely

Ours is a real example. Our tool descriptor declares:

- **`egress`** — the one hostname our agent may reach. Without it, we cannot state a network boundary at all.
- **`auth.check` / `auth.reauth`** — how the platform tests whether a person's credential still works, and what to tell them when it does not.
- **`approvals`** — `deny` on two subcommands that could accept a credential on a command line, and `require_approval` before a command that pulls a full document into the transcript.

Those approval rules are the reason we were comfortable shipping this to other organizations. They are not conveniences; they are the safety argument.

The irony is that we already install our CLI from the skill's first-run step, because of this very issue. So we do not need `install` to work. We would lose the controls while keeping the workaround.

## Suggestion

Reject the parts that genuinely cannot be materialized, and keep the rest:

- `install` on Sprites → error, as you have it
- `sandbox/Dockerfile` on Sprites → error, as you have it
- a descriptor carrying only `egress` / `auth` / `approvals` / `hints` → accept

That keeps the honesty the PR is going for — nobody is misled into thinking a binary will appear — without removing the only place a Sprites deployment can express a network boundary or an approval rule.

If keeping descriptors is not the direction you want, the alternative is a skill-side home for those three fields. Either works for us. Silently having no way to express them is the outcome worth avoiding.

## One smaller thing

The check keys off `config?.target === "fly"`, while the config validator separately requires `"sandbox.backend": "sprites"` to be used only with `target: "fly"`. Deployments currently running `target: docker` with `backend: sprites` — ours is one, and it is what our published guide documents — fail the second check before reaching the first. That combination is worth naming explicitly in the release notes, because the error someone sees will be about `target`, not about the image support that actually changed.

PR #520, opened 2026-08-14
fix(sandbox): make the local backend reachable when core runs in a container

hugo2046:
The `docker` target supports `SANDBOX_BACKEND=local` — `checks.ts` carves out an explicit exception so it does not demand a Fly `sandbox.app`. But on a Linux host that combination cannot work today: every sandbox-backed tool fails with `SANDBOX_BACKEND=local requires a running Docker daemon (is Docker Desktop running?)` while the daemon is perfectly healthy. Peeling it apart turns up three separate walls, all reported through that one misleading message.

**1. The core image has no `docker` CLI.** `createLocalSandbox` shells out via `spawnDockerExec(opts.dockerBin ?? "docker")`, but `deploy/core/Dockerfile` installs only `ca-certificates curl git git-daemon`. `docker version` is ENOENT, and `preflight()` reports it as a stopped daemon.

**2. The mounted socket is unusable by the core process.** The docker backend bind-mounts `/var/run/docker.sock` into core, but core runs as `node` (uid/gid 1000) while the host socket is typically `root:docker` mode 0660. Connecting yields `EACCES` — the mount is there but grants nothing. This does not bite on Docker Desktop, whose socket permissions differ. Fixed by adding `--group-add` with the socket's *actual* gid, read via `statSync`, rather than a hardcoded value; a gid of 0 adds nothing.

**3. The daemon URL assumes core is a host process.** The sandbox publishes its agent port as `127.0.0.1:0:8080` — the *host* loopback — and `daemon()` connects to `http://127.0.0.1:<hostPort>`. When core is itself a container, `127.0.0.1` is its own loopback, so it can never reach the sandbox; the failure changes to `exec daemon never became reachable`. Fixed by teaching core its own container name (`LOCAL_SANDBOX_CORE_CONTAINER`, injected by the docker backend) and, when set, addressing the sandbox as `http://<container>:8080` and joining its network on demand.

**Isolation is unchanged.** Sandboxes keep their own per-scope network; core joins each one rather than sandboxes being attached to core's network, so sandboxes still cannot see each other. Nor does this widen sandbox→core reachability: sandboxes already run with `--add-host=host.docker.internal:host-gateway` and could already reach any host-published port, core included. The host-process path (Docker Desktop) is byte-for-byte unchanged — every new behaviour is behind `if (coreContainer)`.

Also: `network rm` fails while core is still attached, so teardown disconnects first; the two existing `network rm` call sites are consolidated into one `removeNetwork` helper. And the warm-reuse paths ensure the network too — a core recreated by `qm up` otherwise loses its attachment to the networks of sandboxes it later reuses, which the cold-start path alone does not cover.

Tests cover the container-name origin, core joining the network, the disconnect-before-remove ordering, warm rejoin after core recreation, unchanged host-mode behaviour, and gid probing.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/520"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789274832&installation_model_id=19911&pr_number=520&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F520&signature=6548da533a6761cf15a9bca85f777c57a9f67c8655c73f2328fca6a820c74546"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/cli/src/backends/docker.ts
+++ b/cli/src/backends/docker.ts
@@ -1,7 +1,7 @@
 import { httpDeploymentLayerTransport, type DeploymentLayerTransport } from "../deployment-layer.ts";
 
 import { randomBytes } from "node:crypto";
-import { existsSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, mkdtempSync, rmSync, statSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
 import { CliError, bold, die, dim, errMessage, header, note, ok, step, warn } from "../log.ts";
@@ -319,6 +319,7 @@ function serviceEnv(ctx: DockerCtx, service: ServiceName): Record<string, string
     out.SESSION_STORE = "postgres";
     out.RUN_STORE = "postgres";
     out.DATABASE_URL = ctx.databaseUrl;
+    out.LOCAL_SANDBOX_CORE_CONTAINER = cname(ctx, "core");
     if (config.model) out.PI_MODEL = config.model;
     if (config.modelProvider) out.MODEL_PROVIDER = config.modelProvider;
     const layerSubs = existingLayerSubdirs(ctx);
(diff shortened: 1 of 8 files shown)


time-attack:
Thanks for this, @hugo2046 — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/795 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #526, opened 2026-08-14
fix(admin): normalize line endings before hashing CSP script

### LINKED ISSUE #232: CSP hash mismatch blocks the admin shell and auth confirm page on CRLF checkouts
## What happens

The admin shell hangs forever on "Checking your access…" and the browser console shows:

```
Executing inline script violates the following Content Security Policy directive
'script-src 'sha256-<A>''.
Either the 'unsafe-inline' keyword, a hash ('sha256-<B>'), or a nonce is required.
```

The blocked script is the one that resolves the access check, so the page fails
closed at its initial placeholder text.

## Why

The HTML parser normalizes newlines to LF while preprocessing the input stream,
so a browser always hashes the **LF** form of an inline script. Both CSP hashes
are computed over the raw source instead:

- `plugins/admin/src/index.ts:31` — `ADMIN_SCRIPT` is sliced out of `index.html`,
  which `readFileSync` returns verbatim, CRLF included.
- `plugins/auth/src/pages.ts:18` — `CONFIRM_SCRIPT_HASH` is computed over a
  multi-line template literal, so it carries whatever newlines the checkout has.

On an LF checkout the two agree by accident. On a CRLF checkout they can't, and
CSP blocks the script.

Reproduced on Windows with `git config core.autocrlf=true`, which is the default
from the Git for Windows installer. Hashing the admin inline script at `main`
(`ADMIN_BASE_PATH=/admin`) both ways shows the two forms can never agree:

| script bytes     | sha256 (base64)                                |
| ---------------- | ---------------------------------------------- |
| CRLF, as on disk | `8nnxfsr/p6r/aieabKVvy12q73z5l1I4PD+qnXcWfSk=` |
| normalized to LF | `zu5GRQW+lyWKU5O6ueL0RTJ/IzwCrSZCGLta04leSy0=` |

The first is what the header declares on a CRLF checkout; the second is what the
browser computes. The exact pair depends on the current `index.html`, but the
mismatch does not — it follows from the newline form alone.

Nothing rewrites the response in transit: a direct `curl` returns a CRLF body
alongside the CRLF-derived header, which is self-consistent. The mismatch is
introduced by the browser's own normalization, correctly.

`plugins/auth/src/pages.ts` has the same defect. It affects the sign-in confirm
page at `GET /verify`, so the sign-in flow breaks the same way.

## Fix

Normalize CRLF to LF before hashing, in both places. That makes the hash
independent of how the file was checked out.

Worth noting the tests can't catch this — `plugins/admin/test/whoami.test.ts:45`
and `plugins/auth/test/flow.test.ts:477` assert only that a `sha256-` hash is
_present_, not that it matches the served script.

SomSamantray:
## Summary

- Fixes #232: a CRLF checkout (e.g. Windows with `core.autocrlf=true`) breaks the admin shell's inline script, because the server hashes the raw on-disk bytes of the script for its CSP `script-src 'sha256-...'` header, while browsers normalize `\r\n`/`\r` to `\n` before hashing inline script content per spec — the two hashes never match, so the CSP silently blocks the script.
- Extracts the hash computation into an exported `hashAdminScript()` helper (`plugins/admin/src/index.ts`) that normalizes line endings before hashing, so the computed CSP hash always matches what a browser will compute regardless of on-disk line-ending style.

## Test plan

- Added `plugins/admin/test/csp-hash.test.ts`:
  - CRLF and bare-CR variants of a script hash identically to their LF equivalent.
  - Differing content still produces differing hashes (sanity check).
  - The live server's `content-security-policy` response header's `script-src` hash matches `hashAdminScript()` computed over the real on-disk script.
- Verified red-before-fix: temporarily removed the normalization and confirmed exactly the CRLF/CR-specific assertions fail (the LF-only and content-difference assertions still pass), then restored the fix and reran green.
- Full admin plugin suite: `NODE_ENV=test ALLOW_UNSIGNED_TEST_IDENTITY=1 npm test` → 83/83 passing, no regressions.
- `npm run typecheck` → clean.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/526"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789287297&installation_model_id=19911&pr_number=526&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F526&signature=29cbde7c90c11a8c48fd341fa3243899eae4cd8917835b88841e88996680996a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/admin/src/index.ts
+++ b/plugins/admin/src/index.ts
@@ -29,9 +29,15 @@ const BASE_HTML = readFileSync(
   "utf8",
 ).replaceAll("__ADMIN_BASE__", () => ADMIN_BASE_PATH);
 const ADMIN_SCRIPT = BASE_HTML.match(/<script>([\s\S]*?)<\/script>/)?.[1] ?? "";
+// Browsers normalize \r\n and bare \r to \n before hashing an inline <script>'s
+// content for CSP purposes, so the hash here must be computed over the same
+// normalized form or a CRLF checkout produces a mismatch that blocks the script.
+export function hashAdminScript(script: string): string {
+  return createHash("sha256").update(script.replace(/\r\n?/g, "\n")).digest("base64");
+}
 const ADMIN_CSP = [
   "default-src 'self'",
-  `script-src 'sha256-${createHash("sha256").update(ADMIN_SCRIPT).digest("base64")}'`,
+  `script-src 'sha256-${hashAdminScript(ADMIN_SCRIPT)}'`,
   "style-src 'unsafe-inline'",
   "img-src 'self' data:",
   "connect-src 'self'",
--- a/plugins/admin/test/csp-hash.test.ts
+++ b/plugins/admin/test/csp-hash.test.ts
@@ -1,0 +1,74 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createServer, request as httpRequest, type IncomingMessage } from "node:http";
+import type { AddressInfo } from "node:net";
+
+const core = createServer((req: IncomingMessage, res) => {
+  res.writeHead(404, { "content-type": "application/json" });
+  res.end(JSON.stringify({ error: "not_found" }));
+});
+await new Promise<void>((r) => core.listen(0, r));
+const corePort = (core.address() as AddressInfo).port;
+
+process.env.CORE_API_URL = `http://localhost:${corePort}`;
+process.env.CORE_SIGNING_SECRET = "admin-csp-hash-test-secret";
+
+const { server, hashAdminScript } = await import("../src/index.ts");
+
+test("hashAdminScript: LF-only script hashes stably", () => {
+  const script = "console.log('a');\nconsole.log('b');\n";
+  assert.equal(hashAdminScript(script), hashAdminScript(script));
+});
+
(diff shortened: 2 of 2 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/794 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #527, opened 2026-08-14
fix(admin): correct ambient-policy help text to match actual default rule

### LINKED ISSUE #270: Admin's ambient help text describes a channel-size rule the code doesn't implement
# Admin's ambient help text describes a channel-size rule the code doesn't implement

**Version:** CLI 0.1.4, repo at `0f0e0ad`

## What it says

The Admin surface describes the unset/default behavior in three places as depending on how many
people are in the channel (`plugins/admin/public/index.html:4166-4171, 4300-4301`):

```text
Default: on for channels with 8 or fewer members, off for larger ones.
Default (by channel size)
Channels choose per-channel; without a choice, ambient defaults off in channels with more than 8 members.
```

## What the code does

The actual hierarchy is:

1. Org-wide off disables ambient regardless of the per-channel setting.
2. Per-channel `ambientEnabled: false` disables ambient.
3. Per-channel `ambientEnabled: true` enables ambient, subject to the org-wide switch.
4. When the per-channel setting is unset, ambient is enabled only if the channel has a standing
   order or a bot in `action` mode.

Member count is never consulted:

```js
const watchReason =
  (policy?.orders ?? "").trim().length > 0 ||
  Object.values(policy?.bots ?? {}).some((b) => b.mode === "action");

if ((await deps.config?.getOrgAmbientDurable()) === false) {
  ambientOffReason = "ambient is switched off org-wide";
} else if (policy?.ambientEnabled === false) {
  ambientOffReason = "ambient is switched off for this channel";
} else if (policy?.ambientEnabled === undefined && !watchReason) {
  ambientOffReason =
    "ambient defaults off — no standing orders or action-bot triggers here, so only @mentions engage";
}
```

The behavior is also explicit in `test/surface-cache-ambient.test.ts`: no standing orders is the
silent default; standing orders opt in regardless of size; an action-mode bot opts in; explicit
On opts in; and org-wide Off wins over explicit per-channel On.

The Web UI already presents the implemented rule correctly in
`plugins/web-ui/src/ambient-policy.ts:250-257`:

```text
Default (on when standing orders are set)
Default: on only when standing orders (or an action-mode bot) are set below — otherwise mention-only.
```

There is one more stale reference in `test/surface-cache.test.ts:168`, whose assertion message
still says `the size rule decides`.

## How we hit it

A 5-member channel (4 people plus the bot) was logging `decision: ignore` with reason
`ambient defaults off — no standing orders or action-bot triggers here`. Going by the Admin text,
5 ≤ 8 should have meant ambient was already on. Giving the channel a standing order made the
unset/default policy turn itself on. Explicit per-channel On would also opt the channel in, but
becomes unnecessary once a standing order exists.

## Why it matters

The two rules point opposite ways for the common case. An operator reading the Admin copy
concludes a small channel already watches and that a standing order alone is not enough; in
reality the standing order is exactly what enables the unset/default policy. The size-based copy
also disagrees with the Web UI and with the runtime tests.

## Suggested fix

Update all three size-based strings in `plugins/admin/public/index.html` to describe the existing
standing-order/action-bot default, matching the Web UI. The per-channel selector copy should also
make clear that explicit On remains subject to the org-wide switch.

Replace the stale `the size rule decides` assertion message in `test/surface-cache.test.ts` so the
tests no longer preserve the obsolete rule in prose.

If channel size is still intended to control the default, that requires a behavior change plus
updated runtime tests; the current implementation, Web UI copy, and behavior tests all encode the
standing-order/action-bot rule instead.

SomSamantray:
## Summary

The admin panel's "Ambient behavior" help text claimed ambient mode defaults on/off based on channel member count ("on for channels with 8 or fewer members, off for larger ones"). No such logic exists anywhere in the codebase — it's a fictional rule. This PR corrects the copy (in two places) to describe the actual default behavior.

## Root cause

The real default logic lives in `judgeAmbientContainer` (`src/api/app-ambient.ts:89-96`): when a channel's `ambientEnabled` policy is unset ("Default"), ambient is **off** unless the channel has a non-empty standing order (`policy.orders`) or at least one bot configured with `mode: "action"`. There is no member-count check anywhere — confirmed by reading the actual logic and grepping the repo for any member-count-based ambient rule.

## Changes

- `plugins/admin/public/index.html`:
  - Per-channel "Ambient behavior" hint text and the `Default (by channel size)` dropdown option — corrected to describe the standing-order / action-bot condition.
  - Org-wide "Allow ambient behavior" toggle hint — caught during review, since it repeated the same fictional "8 members" claim in a second location.

Copy-only change; `src/api/app-ambient.ts` and all other logic are untouched.

## Validation

- Re-read `judgeAmbientContainer`'s `ambientOffReason` branches line-by-line against the new copy to confirm every claim is directly traceable to the code.
- Grepped the repo for any remaining member-count-based ambient text/logic — none found.
- Ran the changed HTML through Python's `html.parser` to confirm it still parses cleanly.
- No dev-server/browser render was performed (this sandbox has no reachable browser driver and the full dev instance requires Postgres/service setup not available here); the change is plain text content with no new interactive elements.

Fixes #270

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/527"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789287404&installation_model_id=19911&pr_number=527&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F527&signature=f58c77a50b97d03a6218e33f1a63cfd654e988e71f63b50f74c113f723ad482f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/admin/public/index.html
+++ b/plugins/admin/public/index.html
@@ -4196,8 +4196,8 @@
                     <label for="ambient-enabled">Ambient behavior</label>
                     <p class="hint" style="margin: 2px 0 6px">
                       When off, the agent never wakes on overheard messages in this channel — it only responds to direct
-                      @mentions. Default: off, unless this channel has a standing order set below or a bot ledger
-                      entry marked <b>action</b> — either one turns "Default" on.
+                      @mentions. Default: off, unless this channel has a standing order set below or a bot ledger entry
+                      marked <b>action</b> — either one turns "Default" on.
                     </p>
                     <select id="ambient-enabled" style="margin-bottom: 14px">
                       <option value="default">Default (on with standing order or action bot)</option>
@@ -4329,8 +4329,8 @@
                       <span class="setting-copy"
                         ><strong>Allow ambient behavior</strong
                         ><small
-                          >Channels choose per-channel; without a choice, ambient defaults off unless the channel has
-                          a standing order or an action-bot ledger entry.</small
+                          >Channels choose per-channel; without a choice, ambient defaults off unless the channel has a
+                          standing order or an action-bot ledger entry.</small
                         ></span
                       >
                     </label>
--- a/plugins/admin/public/index.html
+++ b/plugins/admin/public/index.html
@@ -4196,10 +4196,11 @@
                     <label for="ambient-enabled">Ambient behavior</label>
                     <p class="hint" style="margin: 2px 0 6px">
                       When off, the agent never wakes on overheard messages in this channel — it only responds to direct
-                      @mentions. Default: on for channels with 8 or fewer members, off for larger ones.
+                      @mentions. Default: off, unless this channel has a standing order set below or a bot ledger
+                      entry marked <b>action</b> — either one turns "Default" on.
                     </p>
                     <select id="ambient-enabled" style="margin-bottom: 14px">
-                      <option value="default">Default (by channel size)</option>
+                      <option value="default">Default (on with standing order or action bot)</option>
                       <option value="on">On</option>
                       <option value="off">Off</option>
                     </select>
--- a/plugins/admin/public/index.html
+++ b/plugins/admin/public/index.html
@@ -4329,8 +4329,8 @@
                       <span class="setting-copy"
                         ><strong>Allow ambient behavior</strong
                         ><small
-                          >Channels choose per-channel; without a choice, ambient defaults off in channels with more
-                          than 8 members.</small
+                          >Channels choose per-channel; without a choice, ambient defaults off unless the channel has
+                          a standing order or an action-bot ledger entry.</small
                         ></span
                       >
                     </label>


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/825 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #528, opened 2026-08-14
fix(cli): require SPRITES_TOKEN for Fly's implicit sandbox default

### LINKED ISSUE #423: qm check passes on a Fly config that can't boot: SPRITES_TOKEN is never requested when sandbox.backend is set but env.core.SANDBOX_BACKEND isn't
### Summary

On `target: "fly"`, the CLI **derives** `SANDBOX_BACKEND` into core's environment from `sandbox.backend`, but the `SPRITES_TOKEN` requirement predicate reads the **declared** `env.core.SANDBOX_BACKEND`. When a config sets `sandbox.backend: "sprites"` (the documented way) and does not *also* set `env.core.SANDBOX_BACKEND`, the two disagree:

- core boots with `SANDBOX_BACKEND=sprites` and demands `SPRITES_TOKEN`
- `qm check` never lists `SPRITES_TOKEN`, and `qm secrets push` never sends it

Result: a green `check`, then a core that crash-loops at boot with `missing or insecure required core secrets: SPRITES_TOKEN`. The failure lands after provisioning, which is the expensive place to find it.

This matters a little more than an ordinary papercut because `check` is the step operators trust to tell them which secrets they must supply.

### Reproduction

Two configs, identical except for one line.

**A — no `env.core.SANDBOX_BACKEND`:**

```jsonc
{
  "contract": 1,
  "orgId": "repro",
  "publicUrl": "https://repro-core.fly.dev",
  "target": "fly",
  "flyOrg": "personal",
  "services": ["core"],
  "region": "iad",
  "modelProvider": "anthropic",
  "env": { "core": {
    "SNAPSHOT_STORE": "s3", "TRANSFER_STORE": "s3",
    "S3_BUCKET": "repro-qm-core", "S3_REGION": "auto"
  } },
  "sandbox": {
    "backend": "sprites",
    "app": "repro-sandboxes",
    "image": "ghcr.io/yc-software/qm/sandbox-base@sha256:<digest>"
  }
}
```

```
$ npx qm check --config repro.jsonc
· required secrets: ANTHROPIC_API_KEY, CAPABILITY_SECRET, CONNECTOR_SECRET_KEY,
  CORE_SIGNING_SECRET, FLY_SANDBOX_API_TOKEN, PORTAL_IDENTITY_SECRET,
  PUBLIC_API_URL, SKILL_SIGNING_SECRET
```

→ **no `SPRITES_TOKEN`.**

**B — add exactly one line, `env.core.SANDBOX_BACKEND: "sprites"`:**

```
· required secrets: …, SKILL_SIGNING_SECRET, SPRITES_TOKEN
```

→ `SPRITES_TOKEN` appears. Nothing else changed.

Observed on `@yc-software/qm@0.1.4`, against `7f2c916`.

### Cause

The requirement predicate (`cli/src/secrets.ts:128-132`):

```ts
{
  name: "SPRITES_TOKEN",
  service: "core",
  required: { when: { kind: "env-equals", service: "core", name: "SANDBOX_BACKEND", value: "sprites" } },
  …
}
```

is evaluated by `conditionMatches` (`cli/src/secrets.ts:379-386`), which resolves the value from the **declared** config plus target defaults:

```ts
const value = (
  config.env[condition.service]?.[condition.name] ?? targetEnvDefault(config, condition.service, condition.name)
)?.trim();
…
return value === condition.value;
```

But the value core actually receives is **derived** (`cli/src/config.ts:228-229`):

```ts
const backend = sb.backend ?? (config.target === "fly" ? "sprites" : undefined);
if (backend) env.SANDBOX_BACKEND = backend;
```

`FLY_TEMPLATE_ENV_DEFAULTS` is `{ core: { HARNESS: "pi" } }` — it does not carry `SANDBOX_BACKEND` — so on Fly the predicate sees `undefined` and never fires.

Note the asymmetry that hides this: the AWS defaults **do** carry `core: { SANDBOX_BACKEND: "aws" }`, so the AWS path resolves correctly. Only Fly is affected, and on Fly the derivation defaults to `sprites` whenever `sandbox.app` is set — so the predicate is wrong precisely when it matters most.

### Suggested fix

The narrow version is to teach the resolver about the derived value, so declared and derived cannot disagree — for example, resolving `SANDBOX_BACKEND` through the same `sb.backend ?? (target === "fly" ? "sprites" : undefined)` expression the deployment uses, rather than reading only `config.env`.

Adding `SANDBOX_BACKEND: "sprites"` to `FLY_TEMPLATE_ENV_DEFAULTS` would also make the reproduction pass, but it is not quite right: the derivation is conditional on `sandbox.app` being present, and a flat default would claim `sprites` for configs that never set up a sandbox.

Happy to open a PR for whichever shape you prefer — I didn't want to guess at the one you'd want to maintain.

### Workaround for anyone hitting this now

Declare it explicitly alongside `sandbox.backend`:

```jsonc
"env": { "core": { "SANDBOX_BACKEND": "sprites" } }
```

That makes the predicate fire, and `check`/`push` then behave correctly.

SomSamantray:
A Fly deployment that left `sandbox.backend` unset (relying on the implicit `"sprites"` default) could pass `qm check` and `qm secrets push` without ever being told `SPRITES_TOKEN` was required, then fail to boot because the sandbox substrate had no token. The two computations had drifted apart: `sandboxCoreEnv` (the actual boot-time env builder) already applied the implicit `"sprites"` default for Fly, but `qm check`'s secret-requirement resolver (`TARGET_ENV_DEFAULTS.fly`) never derived that same default.

This extracts a single shared helper, `flyImplicitSandboxBackend`, that both the boot-time and check-time paths call, so `qm check` now requires `SPRITES_TOKEN` exactly when a deployment will actually need it, and the two can't silently diverge again the way they did here.

Added a regression test covering the four relevant scenarios (implicit default, explicit backend, no sandbox configured, and AWS's own unrelated default staying unaffected), and found three existing test fixtures across the CLI suite that were unknowingly relying on the old bug to avoid provisioning `SPRITES_TOKEN` — updated those to supply it. Full CLI test suite passes on this branch (the small number of remaining failures are pre-existing environment gaps unrelated to this change, present on `main` as well).

Fixes #423.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/528"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789289101&installation_model_id=19911&pr_number=528&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F528&signature=e87d3af30be63d3be9a04dc8baff2bdf6ade9d446f84277eae1f83920ff920e8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/793 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #553, opened 2026-08-16
Relax the docker target for SANDBOX_BACKEND=local (+ optional hardening, + private-network + daemon wiring)

nicwn:
Makes the docker target deployable and working with `SANDBOX_BACKEND=local` (agent computers as co-located Docker containers, no Fly agent-computer app), and fixes three unrelated docker-target gaps that otherwise block the stack from starting or working. `SANDBOX_BACKEND=local` is a core-sanctioned production configuration (the production config test fixture runs on `local`, the production gate names it valid), but no CLI target exposed it for a real deploy. Five commits, all backward-compatible (defaults preserve current behavior):

## 1. Relax docker sandbox.app requirement for local backend

The docker target declares `requiresSandboxApp`, so `qm check`/`qm up` reject a docker deploy with no `sandbox.app` even when `env.core.SANDBOX_BACKEND` is `"local"`. Worse, `qm init --target docker` scaffolds a `sandbox.app` block, which then makes `qm up` throw ("sandbox.app is set but no sandbox layer image is pinned"). The local path now ignores the sandbox block end to end (`qm check` relaxes the requirement, docker-scoped; `sandboxCoreEnv` returns early; `sandboxPinPending`/`sandboxImagePinErrors` skip; `qm doctor` reports "local Docker sandbox: configured"). A shared `localSandboxActive(config)` predicate centralizes the condition.

## 2. Add optional local sandbox hardening controls

Two opt-in env knobs: `LOCAL_SANDBOX_HOST_GATEWAY=0` drops the hardcoded `--add-host=host.docker.internal:host-gateway` flag; `LOCAL_SANDBOX_DISK_GB=N` applies `--storage-opt size=Ng` per container (needs a quota-capable driver). Both default to current behavior.

## 3. Route the docker target's auth broker through a .internal alias

Portal's private-network guard rejects cleartext OIDC endpoints unless the host is loopback/RFC1918 or ends in `.internal`/`.flycast`/`.local`. The docker backend addressed the auth broker as `http://auth:8080` (a bare Docker DNS name), so portal refused to start. Each service container now also gets a `<prefix>-<service>.internal` network-alias, and the auth broker upstream is addressed as `http://<prefix>-auth.internal:8080`.

## 4. Give core access to the host Docker daemon for local sandboxes

`SANDBOX_BACKEND=local` spawns agent computers from core, but the docker target started core without the Docker CLI, socket, or `DOCKER_HOST` — so the sandbox liveness check always failed. Core's image now installs `docker-cli`; when `SANDBOX_BACKEND=local`, the backend bind-mounts `/var/run/docker.sock` into core, adds the socket's group gid so the non-root `node` user can read it, and sets `DOCKER_HOST`. Sprites/fly unchanged.

## 5. Reach the local sandbox from core over the sandbox network, not host localhost

The exec daemon was reached at `http://127.0.0.1:<host-port>`, assuming core is a host process. On the docker target core is a container, so `127.0.0.1` points at core itself and the readiness poll always failed ("exec daemon never became reachable"): the sandbox starts but core can never talk to it. In container mode the backend now sets `QM_CORE_CONTAINER` (core's own container name); the local sandbox omits the host port mapping, connects core to the sandbox's private network (`docker network connect`, reconnected idempotently on cold start so it recovers after a core restart), reaches the daemon by container name (`http://<sandbox>:8080`, resolved by Docker's embedded DNS), and disconnects before removing the network on teardown. **Scope isolation is preserved**: each scope still gets its own `qm-net-<scope>` network, sandboxes cannot reach each other, and core is the only bridge. Host-process mode keeps the existing `127.0.0.1:<host-port>` behavior.

## Verification

- `tsc --noEmit` clean
- `test/config.test.ts`, `test/local-sandbox.test.ts`: 49 pass
- `cli/test/config.test.ts`, `cli/test/check.test.ts`, `cli/test/doctor.test.ts`, `cli/test/auth-broker.test.ts`: 99 pass
- No existing tests changed (one assertion updated in `auth-broker.test.ts` to the new `.internal` address)

Commits 3 and 4 are independent docker-target fixes (affect docker+sprites too); commits 1, 2, and 5 are the local-backend path. Splittable if a reviewer prefers.

### DIFF
--- a/cli/src/backends/docker.ts
+++ b/cli/src/backends/docker.ts
@@ -299,7 +299,7 @@ export function dockerServiceEnv(config: QmConfig, service: ServiceName): Record
       out,
       brokerWiring(service, {
         publicUrl: config.publicUrl,
-        authBaseUrl: "http://auth:8080",
+        authBaseUrl: `http://${dockerPrefix(config)}-auth.internal:8080`,
         ...(config.env.auth?.AUTH_ALLOWED_EMAIL_DOMAIN
           ? { allowedEmailDomain: config.env.auth.AUTH_ALLOWED_EMAIL_DOMAIN }
           : {}),
@@ -406,6 +406,8 @@ function runArgs(ctx: DockerCtx, service: ServiceName, image: string): { args: s
     ctx.network,
     "--network-alias",
     service,
+    "--network-alias",
+    `${cname(ctx, service)}.internal`,
     "--restart",
     "no",
   ];
--- a/cli/src/backends/docker.ts
+++ b/cli/src/backends/docker.ts
@@ -31,7 +31,7 @@ import {
   type LogOpts,
   type ServiceName,
 } from "../services.ts";
-import { dockerBasePort, sandboxCoreEnv, securityScreenEnv, type QmConfig } from "../config.ts";
+import { dockerBasePort, localSandboxActive, sandboxCoreEnv, securityScreenEnv, type QmConfig } from "../config.ts";
 import { discoverPlugins, type ResolvedPlugin } from "../plugins.ts";
 import { computedSecrets, runtimeSecretNames, secretsForService } from "../secrets.ts";
 import { readDeploymentState, withDeploymentLock, writeDeploymentState, type DeploymentState } from "../state.ts";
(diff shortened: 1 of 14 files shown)


nicwn:
Closing this in favor of a proposal in `adrs/` per CONTRIBUTING — see #594. The changes here are a reference implementation of the gaps described in the ADR; happy to share if useful, but I'll let the proposal path run.

PR #558, opened 2026-08-17
web-ui: browse + new-chat palettes and sidebar chrome polish

evebouf:
Adds the **Browse** and **New chat** command palettes with shared grid keyboard navigation, and cleans up the sidebar chrome around them.

## Palettes

`browse.ts`, `new-chat.ts`, `grid-nav.ts` — the two palettes plus the arrow-key grid navigation they share.

Their titles now use the same type as the tile labels beneath them (14px, normal weight, `--foreground`) instead of an 11px semibold muted eyebrow, so the header reads as part of the same system.

## Sidebar

- **Selected chats** use the same tint as hover (`foreground 4%`). The old 9% mix read as too heavy against the light sidebar. Trade-off worth a look: a selected row is now visually identical to a hovered one, so selection only reads when the pointer is elsewhere.
- **Chevron and glyph share one slot.** The folder/channel/group glyph shows at rest; the chevron cross-fades in on hover or keyboard focus; a collapsed group shows the chevron persistently so its state is never ambiguous. The whole header row is still the toggle, so the chevron was only ever an indicator. With the indent rail tightened alongside it, chat titles start 10px further left (29px -> 19px) and project rows with a glyph gain 23px (48px -> 25px).
- **One tooltip, one hover style, everywhere.** Project options, new chat, archive, conversation options, and the sidebar collapse toggle all use the app's floating tooltip rather than the browser's native `title` bubble, and share a single hover treatment (`--secondary` fill, `--foreground` icon) via the base `.session-menu-btn` rule.
- **Accessibility fallout, fixed:** the conversation kebab had only a `title`, so removing it would have left it unlabelled — it now has `aria-label="Options for <chat>"`. The sidebar toggle's `title` was hardcoded to "Hide sidebar" in markup and only corrected later by an imperative updater; label derivation now lives in one `sidebarToggleLabel()` helper, so the tooltip says "Show sidebar" when collapsed.
- The new-chat tooltip drops the project name ("Start a new chat" rather than "Start a new chat in Personal"). The `aria-label` keeps it — a screen reader user has no visual row context to tell those buttons apart.

## Verification

`npm run typecheck` and the full web-ui suite (556 tests) pass; `eslint plugins/web-ui` is clean.

Not yet done, and worth knowing before this lands:

- **No visual demo.** The repo asks for one on front-end changes and I haven't rendered these against a live instance — every claim above is read off the CSS, not off a screenshot. Happy to boot `/dev-instance` and attach before/afters.
- **No fresh-context review pass yet.** Requested separately.

One specific thing to eyeball: `--secondary` is `oklch(0.97 0 0)`, *lighter* than a hovered chat row (~0.945). In the project header that fill sits on an untinted background and reads as a subtle darker chip, but on a chat row — itself tinted while hovered — the icon button may read as a lighter patch instead of a darker one. If it looks inverted, the fix is a row-scoped `color-mix(in srgb, var(--foreground) 9%, transparent)`, same size and shape, preserving the direction of contrast.

## Admin shell (added)

The admin surface sat beside the web UI looking like a different product: a text-only "Agent Admin" string in the top bar, a 258px sidebar with uppercase labels, and its own row metrics. It now uses the same lockup and the same page split, measured rather than eyeballed.

Live measurements at 1440x900, both surfaces on one dev instance (`getBoundingClientRect`):

| | web `/` | admin `/admin/` |
|---|---|---|
| brand row | `x8 y10 267x51` | `x8 y10 267x51` |
| brand mark | `x16.5 y24 22x22` | `x16.5 y24 22x22` |
| brand name | `x46 y26` | `x46 y26` |
| top bar | `x280 y0 1160x56` | `x280 y0 1160x56` |
| nav row height / gap | 33px / 1px | 33px / 1px |

Two non-obvious bits behind that: the web brand row is sized by its 34px collapse toggle, which admin has no equivalent for, so admin needs an explicit `min-height: 51px`; and the mark-to-name gap lives on `.brand-lockup` on the web but on `.brand` in admin, hence 7.5px rather than 9px. Top bar, nav rows and content gutters take the same padding, border and background recipes as `.chat-topbar`, `.navrow` and `.pane`. `.sidebar-title` was a byte-identical duplicate of `.tab-section` and is gone.

Deliberately not matched: admin keeps `--maxw: 1180px` (web list pages use 960) so audit and user tables do not get squeezed, and it keeps identity in the top bar rather than a sidebar footer, since moving it is a navigation change rather than a layout one.

## Sidebar glyph colour (added)

The quick-nav rows read as secondary chrome next to the session rows above them — muted text, icons at 0.85 opacity — while `.session` uses full `--foreground`. Nothing about Home, Create New Chat, Search or Browse is secondary, so `.navrow` takes `--foreground` and the `.quick-nav svg` opacity rule is gone. The collapse toggle drops the `subtle` class that dimmed it below the rail icons it lines up with; its `30px` sizing was dead, as `.icon-btn`'s 34px is declared later and already won.

## Verification (updated)

Booted this worktree with `/dev-instance` and exercised both surfaces in a browser. Admin sidebar, top bar, sessions list and the governance page render correctly at 1440x900 and at 390px; the collapsed rail was checked with all five glyphs. The table above is read off the live DOM, not off the CSS.

562 web-ui tests, 78 admin tests, typecheck, eslint and prettier all pass.

Still missing: no attached screenshots (the CLI cannot upload images to a PR) and no fresh-context review pass on the admin commit. Both worth doing before this lands.

## No shadows (added)

Every surface drops its shadows: web-ui, admin, portal and auth. Elevation was doing work that borders and background tints already do — the hovered refresh button lifted off the page with a two-layer drop shadow while every control around it stayed flat.

Four cases were not simple deletions:

- **Admin focus ring** was `box-shadow: 0 0 0 1px var(--text)` paired with `outline: none`. It becomes a 1px outline of the same colour and width, so keyboard focus is unchanged.
- **Split-canvas drag target** used an inset ring; it becomes `outline` with a negative offset.
- **Monologue row accent bar** was `inset 2px 0 0`. A `border-left` would have shifted that column 2px against every other row, so it is a background gradient instead — which required narrowing the row's `background` shorthands to `background-color`, or the hover state wipes the bar.
- **Dockview** paints floating groups and context menus through `--dv-floating-box-shadow`; nulled once in `.dockview-theme-qm` rather than chased per selector.

Verified by walking the running app and scanning every element's computed `boxShadow` / `textShadow` / `filter` — zero hits on the landing page, the chats list, and with the refresh button hovered. A test in web-ui and admin now fails on any shadow that comes back; portal and auth have no source-reading test to hang that guard on, so their card CSS is unguarded.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/558"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789541486&installation_model_id=19911&pr_number=558&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F558&signature=319887cea2c1f03900c02f6792abc4c2a4444290da75f6a40f59664f79611bd4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


time-attack:
Thanks for this, @evebouf — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/731 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #575, opened 2026-08-18
Docker deploy: give apps a writable /data, and stop reissuing ports after a restart

lmo-selected:
Two defects in the Docker deploy provider. Together they make a locally deployed app that stores anything fail on first run, and then fail differently after a restart. Both are invisible from the outside: the deployment record says `running` while the container has already exited.

## 1. No writable storage

The container gets the snapshot mounted read-only at `/app` and nothing writable anywhere:

```
-v ${version.snapshotDir}:/app:ro
```

So an app that opens a database or writes a file dies on startup. `mkdirSync('/app/data')` fails, and with `{ recursive: true }` it fails the same way, just with `EROFS` instead of `ENOENT`.

The platform already has a contract for this. `aws-deploy-provider.ts` defines `DATA_DIR = "/data"` and `DATA_DB = "/data/app.db"`, and `wiring.ts` warns that an AWS deployment without a data bucket has *"NO durable /data"*. `DeployProfile` even carries a `dataDir` field. The Docker provider never implemented its half — so an app written correctly against that contract works on AWS and crashes locally, which is the wrong way round for the environment people develop in.

Now: a per-deployment directory on the host, mounted read-write at `/data`, `DATA_DIR` in the app's environment, and the path advertised on `profile.dataDir`. The directory lives outside the container, so data survives a redeploy.

## 2. Host ports are allocated in memory

```ts
let nextPort = opts.basePort ?? 9200;
const ports = new Map<string, number>();
```

Every process restart resets the counter to the base. The next `apply` then hands out a port an existing container already holds, and the second container to start fails to bind — after several restarts a set of deployments can all carry the same recorded port.

The allocation is already durable: it is on the deployment record as `endpoint.port`, and `apply` receives the `Deployment`. So this keeps no new state — `apply` prefers the recorded port and advances the counter past it, instead of maintaining a parallel copy in RAM that a restart silently invalidates.

## Tests

`dataRoot` and `mkdir` are injectable alongside the existing `dockerExec`, so both paths are covered without touching a real filesystem or daemon:

- the data directory is created, mounted read-write at `/data`, and `DATA_DIR` is set, while the snapshot stays `:ro`
- a restart, modelled as a fresh provider instance over the same records: a redeploy keeps its recorded port, and a subsequent new deployment does not collide with it

`test/docker-deploy-provider.test.ts` 7/7, `test/aws-deploy-provider.test.ts` 46/46 unaffected, `tsc --noEmit` clean.

## Not fixed here

Nothing reconciles a deployment's recorded `status` against the container's actual state, so a crashed app reports `running` indefinitely and the UI shows it as starting up. That is what made both defects above hard to see, but it needs a reconcile loop rather than a change to this file, so it is left for a separate change.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/575"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789628939&installation_model_id=19911&pr_number=575&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F575&signature=f63415a2b097bd756020549b7fb72c804b072c470359ec01f0de17f023793dac"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/deploy/docker-deploy-provider.ts
+++ b/src/deploy/docker-deploy-provider.ts
@@ -1,26 +1,38 @@
+import { mkdirSync } from "node:fs";
+import { join, resolve } from "node:path";
 import type { Deployment, DeploymentVersion } from "./deploy-store.ts";
 import type { DeployEndpoint, DeployProvider } from "./deploy-provider.ts";
 import { spawnDockerExec, type DockerExec } from "../sandbox/docker-exec.ts";
 
 const APP_PORT = 8080;
+const DATA_DIR = "/data";
 const LEGACY_NETWORK = "agent-deploynet";
 
 export interface DockerDeployProviderOptions {
   image?: string;
   docker?: string;
   basePort?: number;
+  dataRoot?: string;
   dockerExec?: DockerExec;
+  mkdir?: (path: string) => void;
 }
 
 export function createDockerDeployProvider(opts: DockerDeployProviderOptions = {}): DeployProvider {
   const docker = opts.docker ?? "docker";
   const image = opts.image ?? "node:24-alpine";
+  const dataRoot = resolve(opts.dataRoot ?? "./data/deployments-data");
+  const mkdir = opts.mkdir ?? ((path: string) => void mkdirSync(path, { recursive: true }));
   let nextPort = opts.basePort ?? 9200;
   const ports = new Map<string, number>();
   const freed: number[] = [];
-  const allocPort = (n: string): number => {
+  const allocPort = (n: string, assigned?: number): number => {
     const existing = ports.get(n);
     if (existing !== undefined) return existing;
+    if (assigned !== undefined) {
+      ports.set(n, assigned);
+      nextPort = Math.max(nextPort, assigned + 1);
+      return assigned;
+    }
     const port = freed.pop() ?? nextPort++;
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this, @lmo-selected — and sorry for the slow response.

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/789 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #578, opened 2026-08-18
fix(admin): hash the LF-normalized inline script for the CSP source

### LINKED ISSUE #552: Windows: admin SPA blocked by own CSP — inline-script hash computed over CRLF bytes
## Summary

On a Windows checkout (CRLF line endings), the admin plugin's Content-Security-Policy
inline-script hash does not match what browsers compute, so the admin SPA is blocked
by its own CSP (`Refused to execute inline script ... violates the following Content
Security Policy directive`).

Root cause: `plugins/admin/src/index.ts` hashes the raw inline script bytes:

```ts
createHash("sha256").update(ADMIN_SCRIPT).digest("base64")
```

On a Windows checkout `ADMIN_SCRIPT` contains CRLF (`\r\n`). Browsers normalize
inline script text to LF (`\n`) before hashing it (per the HTML spec), so the served
hash covers CRLF bytes while Chromium computes the hash over LF-normalized bytes —
a permanent mismatch on any Windows clone.

Evidence (brute-force confirmed): with the raw CRLF bytes the server hash is
`G6OmXc...`; the LF-normalized script hashes to `WoA/GjO0wOcYzG9SW9axiZFYArelcExznxB9odVmJlw=`
(byte-identical to what Chromium computes), and the SPA boots with 0 console errors
once the server hashes the normalized text.

## Suggested fix (minimal)

Normalize before hashing:

```ts
createHash("sha256").update(ADMIN_SCRIPT.replace(/\r\n/g, "\n")).digest("base64")
```

Alternatively, move the inline script to an external asset so no hash is needed.

## Environment

- OS: Windows (CRLF checkout)
- Browser: Chromium/Edge
- QM pin: b702813382e63e5347b50f542c3e1dcedca052f9

yzxcj797:
## Summary

Fixes #552.

## Root cause (reproduced byte-for-byte)

`plugins/admin/src/index.ts` hashed the inline script exactly as read from `public/index.html`:

```ts
createHash("sha256").update(ADMIN_SCRIPT).digest("base64")
```

The repo stores the file with LF, but on a Windows checkout `core.autocrlf` rewrites the working tree to CRLF, so `readFileSync` returns `\r\n` bytes. Browsers normalize CRLF (and lone CR) to LF in inline scripts **before** hashing for CSP `sha256-` sources — so on any Windows clone the served hash covers bytes the browser will never hash, and the SPA is blocked by its own CSP.

## Fix

Hash the LF-normalized script text via a small exported helper:

```ts
export function cspScriptHash(script: string): string {
  return createHash("sha256").update(script.replace(/\r\n?/g, "\n")).digest("base64");
}
```

`\r\n?` also covers lone `CR`, matching the browser-side normalization. No change to the served HTML itself (CRLF HTML parses fine; only the hash must agree with what the browser computes).

## Tests

New `plugins/admin/test/csp.test.ts`:

- **Unit**: `cspScriptHash` returns the same digest for LF, CRLF, and CR variants of the same script (platform-independent discrimination).
- **End-to-end**: drives the admin server like the existing tests do, fetches `/`, and asserts the served `Content-Security-Policy` contains the hash a browser would compute over the inline script extracted from the served HTML.

Verified on a CRLF Windows checkout — the exact environment from the issue: **both tests fail before the change and pass after it**; the full admin suite is green (81/81).

## Note

`plugins/auth/src/pages.ts` hashes its inline `CONFIRM_SCRIPT` the same way, but that script is a source-level template literal rather than a file read, so it only diverges if the *built* artifact carries CRLF — left untouched here and flagged on the issue for maintainers to consider.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/578"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789640823&installation_model_id=19911&pr_number=578&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F578&signature=49d9b77d89854591e4a8d1968d43672c8d49e2005e8d10f7233116c6b23447bf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/admin/src/index.ts
+++ b/plugins/admin/src/index.ts
@@ -33,9 +33,20 @@ const BASE_HTML = readFileSync(
   "utf8",
 ).replaceAll("__ADMIN_BASE__", () => ADMIN_BASE_PATH);
 const ADMIN_SCRIPT = BASE_HTML.match(/<script>([\s\S]*?)<\/script>/)?.[1] ?? "";
+
+/**
+ * Browsers normalize line endings (CRLF and lone CR to LF) in inline scripts
+ * before hashing them for CSP `sha256-` sources. A Windows checkout reads the
+ * HTML file with CRLF, so hashing the raw bytes yields a hash no browser will
+ * ever compute and the SPA is blocked by its own CSP (#552).
+ */
+export function cspScriptHash(script: string): string {
+  return createHash("sha256").update(script.replace(/\r\n?/g, "\n")).digest("base64");
+}
+
 const ADMIN_CSP = [
   "default-src 'self'",
-  `script-src 'sha256-${createHash("sha256").update(ADMIN_SCRIPT).digest("base64")}'`,
+  `script-src 'sha256-${cspScriptHash(ADMIN_SCRIPT)}'`,
   "style-src 'unsafe-inline'",
   "img-src 'self' data:",
   "connect-src 'self'",
--- a/plugins/admin/test/csp.test.ts
+++ b/plugins/admin/test/csp.test.ts
@@ -1,0 +1,76 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { createHash } from "node:crypto";
+import { createServer, request as httpRequest, type IncomingMessage } from "node:http";
+import type { AddressInfo } from "node:net";
+
+const SCOPES_BODY = { scopes: [{ id: "org:acme", label: "Org", kind: "org" }] };
+
+const core = createServer((req: IncomingMessage, res) => {
+  if (req.method === "GET" && (req.url ?? "").startsWith("/v1/admin/scopes")) {
+    res.writeHead(200, { "content-type": "application/json" });
+    return void res.end(JSON.stringify(SCOPES_BODY));
+  }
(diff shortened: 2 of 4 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/788 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

Issue #577, opened 2026-08-18
SANDBOX_BACKEND=local unusable from published artifacts (sandbox-base has no exec daemon); buildx attestation images misreported as "image not found"


noreplyciy:
Evaluating QM `0.1.4` as a self-hosted deployment (`target: docker`) on Windows + Docker Desktop. Two findings, one blocking and one a genuine bug in the failure path.

## 1. `SANDBOX_BACKEND=local` cannot work from published artifacts

`check.js` requires `sandbox.app` for both the `docker` and `fly` targets, and `config.js` only accepts `"sprites"` or `"aws"` for `sandbox.backend` — but core itself accepts a third value:

```
SANDBOX_BACKEND must be set explicitly in production — use sprites, aws, or local.
```

`local` is reachable by setting `env.core.SANDBOX_BACKEND`, since `docker.js` spreads `config.env[service]` after the computed sandbox env. Core starts fine and chat works.

Agents then can't execute anything. `local-sandbox.ts` runs the sandbox and polls `/health`, but nothing ever listens, because the published image has no exec daemon:

```
$ docker inspect ghcr.io/yc-software/qm/sandbox-base@sha256:52cb44a6e9d1...
Cmd=[sh -c while true; do sleep 3600; done]
```

The same is true of an image produced by `qm sandbox publish` (it inherits that `Cmd`). The image ships `claude`, `codex`, `node`, `gh`, `aws`, but nothing serving the `/health`, `/read`, `/exec` API that `local-sandbox.ts` calls, and `/app/src/sandbox/` in the core image contains only client code. That daemon appears to come from `npm run sandbox:local:build`, which lives in the QM source repo rather than in anything published.

**Question:** is a published local sandbox image (or the build recipe) planned for people deploying from released artifacts rather than the monorepo? Right now `local` is discoverable from core's own error message but unreachable in practice, and `sprites` needs a token from an org that has Sprites enabled — `sprite org auth` returns `sprites not enabled for this organization, email support@fly.io` — so a self-hosted evaluator has no working sandbox backend without going through Fly support first.

## 2. Bug: buildx attestation images make preflight report "image not found"

`local-sandbox.ts` preflight runs:

```
docker image inspect -f '{{.Id}} {{if .Config.Labels}}{{index .Config.Labels "qm.sandbox-fingerprint"}}{{end}}' <image>
```

[...]

PR #580, opened 2026-08-18
fix(sandbox): don't misreport attestation images as missing in local preflight

yzxcj797:
## Summary

Fixes the preflight bug reported in part 2 of #577 (scoped to that; part 1 is a published-artifacts question for maintainers).

## The bug

The local-sandbox preflight ran one probe:

```
docker image inspect -f '{{.Id}} {{if .Config.Labels}}{{index .Config.Labels "qm.sandbox-fingerprint"}}{{end}}' <image>
```

Images built by buildx **with attestations** — the Docker 29 default, and exactly what `qm sandbox publish` emits — inspect as an OCI index whose `Config` has no `Labels` key. The Go template then fails:

```
template parsing error: template: :1:20: executing "" at <.Config.Labels>:
map has no entry for key "Labels"
```

That exits non-zero, so preflight rejected a **present, usable** image as `local sandbox image <image> not found — run npm run sandbox:local:build` — a wrong error that points at rebuilding (the reporter rebuilt twice before checking the exit path itself).

## Fix

Split the probe in two:

1. `docker image inspect -f '{{.Id}}' <image>` decides **existence** — valid for index and manifest images alike.
2. The fingerprint template becomes a second, **tolerated** probe: a non-zero exit means "no fingerprint", which skips the stale-image comparison instead of failing the run.

Fingerprint-labeled images behave exactly as before, including the stale warning.

## Tests

- `test/support/fake-docker.ts` gains the exact attestation behavior: `{{.Id}}` inspects fine, the labels template fails with the real `map has no entry for key "Labels"` error.
- New regression test in `test/local-sandbox.test.ts` asserts provision no longer rejects with the not-found signature. **Verified locally: fails on `main` with precisely that misreport ("image was misreported as missing"), passes with the fix.**

Note: 12 pre-existing tests in this file already fail on Windows checkouts (they exercise the real `agent.mjs` daemon's `/exec`, which needs a POSIX shell) — same count before and after this change; CI on Linux runs them green.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/580"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789643527&installation_model_id=19911&pr_number=580&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F580&signature=83782ce1df251967118ecc69d6474ed797de0b769b433a37225855bccb52da10"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/sandbox/local-sandbox.ts
+++ b/src/sandbox/local-sandbox.ts
@@ -112,21 +112,28 @@ export function createLocalSandbox(workspace: WorkspaceStore, opts: LocalSandbox
         preflightDone = undefined;
         throw new Error("SANDBOX_BACKEND=local requires a running Docker daemon (is Docker Desktop running?)");
       }
-      const img = await dexec([
-        "image",
-        "inspect",
-        "-f",
-        `{{.Id}} {{if .Config.Labels}}{{index .Config.Labels "${FINGERPRINT_LABEL}"}}{{end}}`,
-        image,
-      ]);
+      // Existence and label are separate probes: images built by buildx with
+      // attestations (the Docker 29 default, and what `qm sandbox publish`
+      // emits) inspect as an OCI index whose Config has no Labels key, which
+      // fails the fingerprint template even though the image is present and
+      // usable. A template failure must not be reported as a missing image.
+      const img = await dexec(["image", "inspect", "-f", "{{.Id}}", image]);
       if (img.code !== 0) {
         preflightDone = undefined;
         throw new Error(`local sandbox image ${image} not found — ${BUILD_HINT}`);
       }
-      const [imageId = "", labeled = ""] = img.stdout.trim().split(/\s+/);
+      const labeled = await dexec([
+        "image",
+        "inspect",
+        "-f",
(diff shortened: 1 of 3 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/786 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #583, opened 2026-08-18
fix(cli): probe the configured provider endpoint in doctor

### LINKED ISSUE #525: Docker doctor ignores configured OpenAI-compatible endpoint
When `OPENAI_BASE_URL` is configured, runtime/Admin endpoint resolution uses that compatible endpoint, but the Docker doctor still probes the official OpenAI `/v1/models` URL. This can report a false credential failure for a key that is valid only at the configured endpoint.

Current behavior is in `cli/src/backends/doctor.ts`; the Admin model-provider route already resolves the configured provider base URL.

Expected: doctor should probe the normalized configured OpenAI endpoint when present, while preserving current loopback/private HTTP support and rejecting credential/query/fragment URL components consistently with the shared endpoint parser.

This is separate from the Codex app-server routing behavior reported in #522: a successful `/models` probe validates the configured credential endpoint, but does not prove Responses API compatibility.

yzxcj797:
## Summary

Fixes #525.

## The bug

`cli/src/backends/doctor.ts` hardcoded the official endpoints in `MODEL_PROVIDER_PROBES`, so with `OPENAI_BASE_URL` pointing at an OpenAI-compatible gateway the doctor probed `api.openai.com/v1/models` and reported a **false credential failure** for a key that is only valid at the configured endpoint — while runtime/Admin endpoint resolution (`src/api/routes/admin/model-providers.ts` → the shared `provider-endpoints` module) used the override.

## Fix

The doctor now resolves the probe URL the way the Admin route does — `override ?? default` — reading `OPENAI_BASE_URL` / `ANTHROPIC_BASE_URL` / `OPENROUTER_BASE_URL` from the **same secrets map** the API keys are read from, so a staged/local `.env` override is honored consistently.

The CLI package doesn't import core modules, so the shared parser's validation rules are mirrored in a new exported `modelProviderProbeUrl(provider, envValue)`:

- http(s) only,
- no embedded credentials, no query string, no fragment,
- trailing slashes trimmed,
- **loopback and private HTTP hosts stay allowed** (gateway deployments on internal networks are the point of the override).

Probe shapes: `openai` → `<base>/models` (base default `https://api.openai.com/v1`), `anthropic` → `<base>/v1/models?limit=1`, `openrouter` → `<base>/key`.

## Tests

Three new unit tests over `modelProviderProbeUrl`:

- no-override defaults resolve to the official endpoints (whitespace-only values treated as unset);
- the override resolves the exact false-failure case from the issue, including trailing-slash trimming and a loopback host;
- each parser rejection (non-HTTP scheme, credentials, query, fragment, unparseable) throws the same class of `CliError` as the shared parser's messages.

Local run: 21 passing (18 pre-existing + 3 new); the 5 pre-existing Fly-doctor failures on this Windows checkout are unchanged by this PR and pass on CI.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/583"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789644453&installation_model_id=19911&pr_number=583&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F583&signature=e3a89627539891084496f788061597e6f8736c5fa92c236c0b7345e75f6f95aa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/cli/src/backends/doctor.ts
+++ b/cli/src/backends/doctor.ts
@@ -214,26 +214,70 @@ async function baseModelCheck(config: QmConfig, secrets: Map<string, string>): P
     warn(`${name} is not available locally — skipping the live ${provider} check`);
     return;
   }
-  await modelProviderCheck(provider, key);
+  await modelProviderCheck(provider, key, secrets);
   step(`base model provider ${provider}: ${name} accepted`);
 }
 
 const MODEL_PROVIDER_PROBES: Readonly<
-  Record<ModelProvider, { url: string; headers: (key: string) => Record<string, string> }>
+  Record<ModelProvider, { baseUrl: string; path: string; headers: (key: string) => Record<string, string> }>
 > = {
   anthropic: {
-    url: "https://api.anthropic.com/v1/models?limit=1",
+    baseUrl: "https://api.anthropic.com",
+    path: "/v1/models?limit=1",
     headers: (key) => ({ "x-api-key": key, "anthropic-version": "2023-06-01" }),
   },
-  openai: { url: "https://api.openai.com/v1/models", headers: (key) => ({ authorization: `Bearer ${key}` }) },
-  openrouter: { url: "https://openrouter.ai/api/v1/key", headers: (key) => ({ authorization: `Bearer ${key}` }) },
+  openai: {
+    baseUrl: "https://api.openai.com/v1",
+    path: "/models",
+    headers: (key) => ({ authorization: `Bearer ${key}` }),
+  },
+  openrouter: {
+    baseUrl: "https://openrouter.ai/api/v1",
+    path: "/key",
+    headers: (key) => ({ authorization: `Bearer ${key}` }),
+  },
 };
 
-async function modelProviderCheck(provider: ModelProvider, apiKey: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/784 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #593, opened 2026-08-18
fix(admin): show the harness/model card at every accepted scope

### LINKED ISSUE #589: Admin UI only offers the harness/model control at org scope, though the resource accepts any scope
# Admin UI only offers the harness/model control at org scope, though the resource accepts any scope

The `runtime` resource is declared `target: "any"` in
`src/api/routes/admin-resources.ts`, and it works on a channel scope. Against my
own deployment:

```
GET  /v1/admin/scopes/channel:C0B…            runtime = {"harnessId":"pi","modelId":"kimi-k3"}
PUT  /v1/admin/scopes/channel:C0B…/runtime    {"harnessId":"pi","modelId":"deepseek-v4-flash"} → 200
GET  /v1/admin/scopes/channel:C0B…            runtime = {"harnessId":"pi","modelId":"deepseek-v4-flash"}
```

It isn't alone: the `base-model` resource (`admin-resources.ts`, same file) is
also `target: "any"` and `clearable` — an enum form writing the same storage,
restricted to the registry's built-in models. Custom models like
`deepseek-v4-flash` only go in through `runtime`, which is the path above.

The admin UI hides both everywhere except the org:

```js
const showBaseModel = scope.startsWith("org:") && opts.length > 0 && dflt && "runtime" in r.data;
$("card-base-model").classList.toggle("hidden", !showBaseModel);
```

So an admin who wants a particular channel or project on a particular model has
to call the API by hand or write `base_model_configs` directly. Both work;
neither is something I'd hand to a customer's admin.

## Why per-scope is the interesting case

The Web UI picker isn't a substitute. That's the user choosing per conversation
from an allowlist; this is an admin deciding what a room runs on, which is a
different question with a different answer.

Different rooms have genuinely different requirements, and models differ in ways
that matter. Measuring twelve models against qm
([harness and data](https://github.com/ianTPE/qm-model-eval)), I planted "the
office wifi password is …" in a memory notebook and asked the agent to remember
an unrelated second fact. `claude-opus-5`, `deepseek-v4-flash` and `qwen3.8-max`
kept the password and stored an "internal code" alongside it, every run.
`gpt-5.6-sol` deleted the password, declined the code, and said why — no
credentials in long-term memory.

That was an accidental finding from a badly-chosen probe and it deserves a
deliberate test before anyone leans on it. But it's the shape of the thing: given
that a qm memory is readable from every other conversation in the org (default
recall policy is `visible`, which spans every workspace layer including the org —
`src/memory/policy.ts`), an admin may well want the finance channel on a model
with that instinct and the rest of the org on something cheaper. Among the nine
models that passed every task in my set, list prices span about 25×, so "same
model everywhere" is also an expensive default.

The backend already supports all of this. It's the one condition in the UI that
doesn't.

## Suggestion

Drop the `org:` gate and drive visibility off the manifest instead of the
response body. One wrinkle worth naming: the `"runtime" in r.data` conjunct in
that condition looks like a signal but isn't — `getScopeConfig`
(`src/api/routes/admin/scope-config.ts`) writes every resource with a `readKey`
into the response unconditionally regardless of scope, so the key is present for
channels and projects too. The authoritative signal is the resource manifest
(`adminResourceManifest()` in `admin-resources.ts`), which already exposes
`target: "org" | "any"` per resource — the same field the backend enforces
against. Inheritance is already handled: both resources are `clearable`, and
`{"inherit": true}` clears a scope override back to the org default, so a
per-scope control has somewhere sensible to reset to.

yzxcj797:
## Summary

Fixes #589.

## Root cause

Purely a UI gate: the `runtime` (and `base-model`) resources are declared `target: "any"` in `admin-resources.ts`, and `scope-config.ts` already returns `baseModelOptions` / `modelsByHarness` / `runtime` for **every** scope — which is exactly why the issue's API round-trip works on a channel while the card stays hidden. The one-line culprit:

```js
const showBaseModel = scope.startsWith("org:") && opts.length > 0 && dflt && "runtime" in r.data;
```

## Fix

Drop the `org:` prefix from `showBaseModel` — the harness/model card now renders at every scope the resource accepts, and per-channel/per-project model selection happens through the same form as org-level.

Deliberately untouched, because their backends are genuinely org-scoped:

- `approved-harnesses` (`target: "org"`, org-wide by design),
- the ambient/external-Slack/fast-mode/header-pin cards (all org-scoped resources in the same block).

No backend change needed.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/593"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789665651&installation_model_id=19911&pr_number=593&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F593&signature=fd4a873caa57350d77d34999659490c8fa39b1fd91dada6c164d204fd0485a6a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/admin/public/index.html
+++ b/plugins/admin/public/index.html
@@ -6413,7 +6413,11 @@
           Array.isArray(r.data.approvedHarnesses) && r.data.approvedHarnesses.length
             ? r.data.approvedHarnesses
             : [harnessDefault];
-        const showBaseModel = scope.startsWith("org:") && opts.length > 0 && dflt && "runtime" in r.data;
+        // The runtime/base-model resources accept any scope (admin-resources.ts
+        // target: "any"); the org: prefix here hid them for channels and
+        // projects, forcing per-scope model choices through hand-rolled API
+        // calls (#589).
+        const showBaseModel = opts.length > 0 && dflt && "runtime" in r.data;
         $("card-base-model").classList.toggle("hidden", !showBaseModel);
         if (showBaseModel) {
           const current = r.data.runtime || { harnessId: harnessDefault, modelId: r.data.baseModel || dflt };


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/821 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

Issue #560, opened 2026-08-17
Feature request: an HMAC-reachable route to read session content (transcript / tool calls)


QueRikii:
**Summary.** Routes that expose session content — `GET /v1/sessions/:id`, `GET /v1/sessions/:id/approvals` — are user-scoped, so a `auth: "source"` integration gets `401 portal identity required` even though it holds `CORE_SIGNING_SECRET`. There is no source-auth path to read what a session actually contains. Two concrete consequences:

**1. A human approving a Strict-posture gate cannot be shown what they are approving.**

Under Strict, `withToolApprovalGate` (`src/harness/pi-tools.ts:2433-2441`) fires on the tool *name* before execution and pushes:

```ts
ref.pendingApprovals?.push({
  command: tool.name,
  reason: STRICT_TOOL_APPROVAL_REASON,
  kind: "approval",
  approvalKey: `tool:${tool.name}`,
});
```

`params` is in scope on that line and is not included. `recordCall` immediately below logs `{tool, blocked, reason}` — also without params. So an integration building an approval queue can tell a human *"the agent wants to use `execute`"* but never *"…to run `rm -rf /var/data`"*. The arguments exist only in the model's tool-use block in the transcript, which is precisely what source-auth cannot read.

That seems worth surfacing on its own terms: an approval UI that shows the category rather than the action is weaker than it looks, and every integrator hits this identically. (If the arguments were simply added to the pending-approval record, that would solve this half outright and would be a smaller change than a new route — genuinely happy with either.)

**2. An integration cannot render what the agent produced.**

Anything showing an agent's work — a live view, an audit surface, a "what did this run actually do" panel — needs the transcript. Today only a portal-identity holder can read it, so an integration must either hold an impersonation-capable credential or show nothing.

**Proposed shape.**

```
GET /v1/sessions/:id/content?since=&limit=      auth: "source"
→ { entries: [ { seq, type, at, tool?, args?, text? } ], nextCursor }
```

Or, more narrowly and probably more useful as a first step: include the tool-call arguments in `pendingApprovals[]` on `POST /v1/turns`, which solves consequence 1 without exposing whole transcripts.

**Why source-auth is defensible here.** A source-auth peer already holds `CORE_SIGNING_SECRET`, can post turns, and can resume any approval — it can already *cause* the actions whose record it is being denied. Withholding read while granting write is the awkward half of the current split. If org-scoping is wanted, an `orgId` filter would still solve it.

[...]

PR #615, opened 2026-08-20
fix(harness): strict-posture approval records carry the gated call's arguments

yzxcj797:
Fixes the narrower half of #560 — the reporter's explicitly preferred first step ("If the arguments were simply added to the pending-approval record, that would solve this half outright and would be a smaller change than a new route"). The HMAC-reachable session-content route is deliberately **not** included: widening source-auth read access to transcripts is a trust-model decision only the maintainers can make, and this half stands on its own.

## What changes

Under Strict posture, `withToolApprovalGate` fired on the tool *name* and recorded only that name — `pendingApprovals` got `{command: "execute"}`, and the transcript's `tool_call`/`tool_result` records got `{tool, blocked, reason}`. An approver (human or integration) could see the category but never the action: *"the agent wants to use `execute`"*, never *"…to run `rm -rf /var/data`"*. The arguments lived only in the model's tool-use block in the transcript — precisely what a `auth: "source"` integration cannot read.

The gate now renders the call's arguments:

1. **`pendingApprovals` gains `summary`** — the field that already exists on `PendingApproval`/`PendingApprovalRecord` for exactly this purpose (the LLM-generated exec approval summaries). Here it's deterministic: JSON for object params, verbatim for strings, capped at 300 chars (the same cap the orchestrator's LLM summaries use, so a huge payload can't flood the approval surface), and **omitted entirely** for argless or unrenderable calls.
2. **The persisted `tool_call`/`tool_result` records gain the same bounded `args` rendering**, so the transcript audit shows *what* was blocked, not just that something was.
3. **The orchestrator prefers a producer-provided summary over generating an LLM one** (`pa.summary ?? await approvalSummary(...)`): for tool-gate entries the LLM pass would only ever see the bare tool name, so the deterministic rendering is strictly more informative — and skipping the round-trip saves the model call. No existing producer sets `summary`, so exec-path approvals behave exactly as before.

`HarnessTurnResult`/`ToolContextRef` pending-approval item types gain the optional `summary` field; the public API types already had it.

## Testing

Two new tests in `test/pi-tools.test.ts`:

- `a strict-posture gate record carries the tool call's arguments` — `pendingApprovals[0].summary` matches the gated command, and both persisted records carry the same `args`. **Fails on `main`** (no summary/args anywhere).
- `a strict-posture gate renders huge arguments bounded and argless calls without a summary` — a 5,000-char command is capped at the 300-char bound with an ellipsis; an argless gated call produces no `summary` field at all. The bounded half **fails on `main`** (no rendering exists).

Full file: 71/71 pass with the fix, 69 pass + the 2 new failing on unpatched `main`. `orchestrator`/approval test files: identical results with and without the change in this environment (84 pass / 60 pre-existing environment failures in both runs), zero regressions.

## What an integrator gets

With this, an approval-queue integration building on `POST /v1/turns` (source-auth) can render *"the agent wants to run `execute` with `{"command":"rm -rf /var/data"}`"* — a decidable ask — instead of the tool name alone. The approver already controls execution of that exact call; showing them the call's arguments leaks nothing they can't already cause.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/615"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789786491&installation_model_id=19911&pr_number=615&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F615&signature=5494ed688298a57f981a6d969a8773e76b8f99576e0c6ca2405564ca8545ff9a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -2898,7 +2898,10 @@ export function createOrchestrator(deps: OrchestratorDeps): Orchestrator {
             const blocks = approvalBlocksInput(pa.kind, outcome);
             const command = pa.command;
             const requestId = commandApprovalId(session.id, command);
-            const summary = await approvalSummary(scopeId, command, pa.reason, pa.purpose);
+            // A producer-provided summary (the strict tool gate renders the
+            // call's arguments deterministically) needs no LLM pass and is
+            // more specific than one generated from the bare tool name.
+            const summary = pa.summary ?? (await approvalSummary(scopeId, command, pa.reason, pa.purpose));
             prepared.push({
               requestId,
               record: {
--- a/src/harness/harness.ts
+++ b/src/harness/harness.ts
@@ -105,6 +105,7 @@ export interface HarnessTurnResult {
     matched?: string;
     purpose?: string;
     approvalKey?: string;
+    summary?: string;
   }>;
   pausedOnApproval?: boolean;
   modelCalls?: number;
--- a/src/harness/pi-tools.ts
+++ b/src/harness/pi-tools.ts
@@ -38,6 +38,7 @@ export interface ToolContextRef {
     matched?: string;
     purpose?: string;
     approvalKey?: string;
+    summary?: string;
   }>;
   pausedOnApproval?: boolean;
   emit?: (entry: { type: EntryType; payload: unknown; scopeLabel: ScopeId }) => void | Promise<unknown>;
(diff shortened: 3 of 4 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/776 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #616, opened 2026-08-20
feat(governance): resolve memory policy per scope like the other governance settings

### LINKED ISSUE #559: Memory policy is the only governance setting that isn't per-scope
Memory policy is the one governance setting that doesn't resolve per scope.

`admin-resources.ts` exposes ambient-policy, egress, security-posture,
command-policy, approval-grant-modes, soul, runtime and the rest as scope-keyed
resources — 21 of the 23 take a scope argument, and the Governance page renders
them for whichever scope you're on. Memory isn't among them. `memoryPolicy` is
built once in `wiring.ts` from `MEMORY_CAPTURE` / `MEMORY_RECALL` in `config.ts`,
and the orchestrator reads `deps.memoryPolicy ?? DEFAULT_MEMORY_POLICY` with no
scope in the lookup. So it's the whole deployment or nothing.

The case I ran into thinking about it: a channel where HR discusses a specific
employee. I'd want capture off in that one channel and unchanged everywhere
else. Today the only lever is not inviting the bot — which is probably the right
answer for that particular channel anyway, but it means there is no middle
setting for "the agent may take part and must not remember."

The org-memory path is what makes it sharp rather than cosmetic. Every scope
mounts the org scope as a read-only layer, and with recall at its default
(`visible`) that layer is read in every scope, including the subject's own DM.
An org admin's turn is granted `orgWrite` to it — that grant is decided by
`actorIsOrgAdmin`, so it's identity, not model behaviour. What model behaviour
decides is *what gets written there*, and there's no scope-level guardrail on
that at all. Nothing has written to `org:` in two weeks of use here, so this is
a shape I'm reporting rather than an incident.

Memory seems like the setting most worth having per scope, given it's the one
that travels between conversations.

yzxcj797:
Fixes #559.

## What this does

Memory was the one governance setting without scope resolution: 21 of the 23 admin resources take a scope, but `memoryPolicy` was built once in `wiring.ts` from `MEMORY_CAPTURE`/`MEMORY_RECALL` and every turn read that single deployment-wide value. The only lever for *"the agent may take part in this channel and must not remember it"* was not inviting the bot. This PR makes memory policy scope-keyed with the same shape as `security-posture`:

- **Config store** — a persisted `memory_policies` map with `getMemoryPolicyDurable` / `setMemoryPolicy` / `clearMemoryPolicy`. The **org scope is the privacy floor** (the deployment env config remains its default when nothing is stored); a narrower scope may **tighten** either axis — `capture` to `off`, `recall` by tightness rank `off < writable < visible` — but can never loosen below the floor. The composed effective value is stored on write, mirroring `setSecurityPosture` (reads stay single-lookup, and a stored scope value can never be looser than the floor even if the floor tightens later).
- **Resolution** — the per-turn `Resolution` now carries the effective `memoryPolicy` for the resolved scope (alongside `commandPolicy`/`securityPolicy`/`approvalGrantModes`); the orchestrator's three capture/recall sites read the per-turn policy, falling back to the deployment-wide `deps.memoryPolicy` when a resolution was built without one — so every existing resolution in tests and embedded paths behaves exactly as before.
- **Admin surface** — a `memory-policy` resource (`target: any`) renders in the Governance page for whichever scope you're on, validates `{ recall, capture }` fully (unknown values → 400, both-axes-required matches `approval-grant-modes`), and appears automatically in `/v1/admin/resources`.

The env knobs are unchanged: a deployment that sets only `MEMORY_CAPTURE`/`MEMORY_RECALL` observes today's behavior in every scope; the floor just becomes overridable-tighter per scope.

## Composition semantics

Same tighten-only rule the issue's framing implies and the repo already uses elsewhere ("The org value is a minimum; narrower scopes may tighten it but cannot weaken it" — security-posture's label). Concretely, `composeMemoryPolicy(floor, scopeValue)`:

| org floor | scope sets | effective |
| --- | --- | --- |
| capture `writable`, recall `visible` (default) | capture `off` | capture `off`, recall `visible` |
| capture `writable`, recall `visible` | recall `writable` | capture `writable`, recall `writable` |
| capture `off` (org-wide off) | capture `writable` (attempt to loosen) | capture `off` — floor wins |

The org-memory concern from the issue is covered by the floor direction: an org that wants `org:` writes bounded sets the org floor to `capture: writable, recall: writable` and no channel can widen recall past the writable scope.

## Tests (`test/memory-policy-scope.test.ts`)

- `composeMemoryPolicy` — axis independence, floor-cannot-loosen, recall ranking, identity.
- Config store — deployment-default fallback, org floor set, channel loosen-rejected, channel tighten accepted, clear returns to floor.
- **The issue's exact scenario end-to-end** — one channel with `capture: off` does not retain or recall a stated fact ("remember that I own the billing service"), while an untouched channel captures and recalls its fact in the same app instance.

All three fail on unpatched `main` (the API does not exist there). Regression run over `memory.test.ts` + `memory-policy-scope.test.ts` + `config-store-org.test.ts` + `admin-resources.test.ts` + resolution suites: 39/39 pass, zero regressions; full `tsc --noEmit` clean.

## Notes for reviewers

- The resource requires both axes in the PUT body (`{ recall, capture }`) rather than partial updates — consistent with `approval-grant-modes`, and the UI always edits them as a pair.
- `clearMemoryPolicy` exists on the store for symmetry with the other scoped settings; I did not mark the resource `clearable` in the manifest since the floor (org/env) remains after a clear — happy to wire the clear endpoint if you want parity with `base-model`.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/616"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789787435&installation_model_id=19911&pr_number=616&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F616&signature=be27052da2c2ffb24b841fd1132dd31427459f97b436ac7ec7141cfddffb4593"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/api/routes/admin-resources.ts
+++ b/src/api/routes/admin-resources.ts
@@ -32,6 +32,7 @@ import { encodeRef, serviceCredRef } from "../../acl/resource-ref.ts";
 import { audit } from "./shared.ts";
 import { errMessage } from "../../util/errors.ts";
 import { parseSecurityPosture, SECURITY_POSTURES, type SecurityPosture } from "../../security/security-posture.ts";
+import { DEFAULT_MEMORY_POLICY, type MemoryPolicy } from "../../memory/policy.ts";
 import type { ApprovalGrantModes } from "../../types.ts";
 import { parseEgressPolicy } from "../../resolution/egress-policy.ts";
 import { DEVICE_FLOW_CUTOVER_MODES, type DeviceFlowCutoverMode } from "../../credentials/device-flow-cutover.ts";
@@ -123,6 +124,36 @@ export const ADMIN_RESOURCES: readonly AdminResource[] = [
         return { value: { session: b.session, always: b.always } };
       },
       (deps, scope, modes) => deps.config!.setApprovalGrantModes(scope, modes),
+    ),
+  },
+  {
+    id: "memory-policy",
+    kind: "custom",
+    target: "any",
+    label:
+      "Memory capture and recall for this scope. The org value is a privacy floor: narrower scopes may tighten (capture off, recall restricted) but cannot loosen it.",
+    readKey: "memoryPolicy",
+    get: (deps, scope) => deps.config!.getMemoryPolicyDurable(scope),
+    apply: generic<MemoryPolicy>(
+      (body) => {
+        const b = body as { recall?: unknown; capture?: unknown };
+        if (b.recall !== undefined && !["off", "writable", "visible"].includes(String(b.recall))) {
+          return { error: 'memory-policy recall must be "off" | "writable" | "visible"' };
+        }
+        if (b.capture !== undefined && !["off", "writable"].includes(String(b.capture))) {
+          return { error: 'memory-policy capture must be "off" | "writable"' };
+        }
+        if (b.recall === undefined && b.capture === undefined) {
+          return { error: "memory-policy requires { recall, capture }" };
+        }
+        return {
+          value: {
+            recall: (b.recall as MemoryPolicy["recall"]) ?? DEFAULT_MEMORY_POLICY.recall,
+            capture: (b.capture as MemoryPolicy["capture"]) ?? DEFAULT_MEMORY_POLICY.capture,
+          },
+        };
+      },
+      (deps, scope, policy) => deps.config!.setMemoryPolicy(scope, policy),
     ),
   },
   {
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -2898,7 +2898,10 @@ export function createOrchestrator(deps: OrchestratorDeps): Orchestrator {
             const blocks = approvalBlocksInput(pa.kind, outcome);
             const command = pa.command;
             const requestId = commandApprovalId(session.id, command);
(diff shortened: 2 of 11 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/775 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #617, opened 2026-08-20
feat(files): collapse viewer-facing file lists to one row per document

### LINKED ISSUE #601: Files lists one document once per turn that touches it, so a project's file view grows without bound
# Files lists one document once per turn that touches it, so a project's file view grows without bound

A file artifact's identity is derived from the turn that registered it, and
nothing collapses rows across turns. The bytes are deduplicated correctly; the
listing is not. In a project where an agent hands the same document back a few
times, the Files view fills with entries that are byte-identical, same-named, and
indistinguishable at a glance.

(Filenames below are anonymised — this came out of a finance project on my own
deployment. Structure and counts are verbatim.)

## What I saw

Four documents, ten rows, one project scope:

| time | actor | direction | rows |
|---|---|---|---|
| 17:08:13 | `sales@…` | `in` | 2 |
| 17:22:14 | `finance@…` | `in` | 4 |
| 17:22:51 | agent reply carried them | `out` | 4 |

The two documents both people contributed have **three** rows each; the two only
one person contributed have **two**. Every row for a given document has the same
`sha256`, the same `blob_key`, a different `id`, and a different `path`.

The byte layer is fine — `keyFor` in `src/files/durable-byte-store.ts:42` is
`files/${sha256}`, and the store `stat`s before writing (:83). On disk there are
exactly four blobs for four documents, and the two written at 17:08 still carry
that mtime after the 17:22 re-registration. Nothing is stored twice. It is only
the artifact rows, and the view built from them, that multiply.

## Why it grows

`src/core/orchestrator.ts:1609` seeds the artifact id with the run:

```ts
const fileRegistration: ArtifactRegistration = {
  store: deps.files,
  ownerScopeId: toScopeId("personal", actor.id),
  createdBy: actor.id,
  createdInScope: scopeId,
  seed: input.runId ?? `${session.id}:${Date.now()}`,
```

```ts
// src/files/file-artifact-store.ts:75
export function fileArtifactId(seed: string, direction: FileDirection, batchIndex: number): string {
  return createHash("sha256").update(`${seed}:${direction}:${batchIndex}`).digest("hex").slice(0, 32);
}
```

and the insert dedupes on that id alone:

```sql
-- src/files/postgres-file-artifact-store.ts:89
ON CONFLICT (id) DO NOTHING
```

So the idempotency key is the *turn*. Retrying within a run is safe, which is
presumably the intent; but two turns that carry the same bytes are two rows by
construction. An agent that attaches the same spreadsheet in ten replies leaves
ten rows. This is not a fixed +2 — it is one row per document per turn that
touches it, and it only accumulates.

This is also a property of the registration path, not of artifact ids in
general. Every conversation attachment — inbound via `materializeInbound`,
outbound via `collectOutbound`/`collectNamedOutbound`
(`src/core/attachments.ts`) — flows through the same per-turn seed. The one
counter-example in the repo is the write tool's share path
(`src/tools/primitives.ts:612`), which registers with a *stable* id
(`fileArtifactId(\`${writableScopeId}:${path}\`, "out", 0)`) and the writable
scope as owner — so re-sharing an unchanged file across turns collapses to one
row via the same `ON CONFLICT`. Cross-turn deduplication by id choice is
already an established pattern here, just not on the conversation path where
the multiplication actually happens.

The listing has no collapsing step to compensate:

```sql
-- src/files/postgres-file-artifact-store.ts:141
SELECT * FROM file_artifacts
  WHERE ... created_in_scope = $n ...
  ORDER BY created_at DESC, id DESC
```

`sha256` is stored on every row and is already the blob key, so the information
needed to group is present and unused. The indexes that exist are
`(owner_scope_id, path)` (non-unique, and `path` embeds the id, so it varies
with the same turn seed that made the row distinct in the first place) and a
partial `(created_in_scope, created_at DESC, id DESC) WHERE enabled = TRUE`
(:39-40) — the latter already serves the scope listing's ordering; what's
missing is the grouping key.

## A second effect, same root

`src/core/orchestrator.ts:1606`:

```ts
ownerScopeId: toScopeId("personal", actor.id),
```

A document surfaced inside a project is owned by whoever's turn produced it. In
my ten rows the same document is owned by two different people. So a project's
files aren't the project's; they're a union of personal artifacts that happen to
share a `created_in_scope`. That is defensible for ACL purposes, but it means the
duplicate rows also disagree about ownership, and there's no row that represents
"this document, in this project."

## Why this matters more than it looks

The scope where I hit it was finance. Three identically-named copies of a
reimbursement schedule, with no visible marker of which arrived from whom and
which the agent emitted, is not a cosmetic problem — picking the wrong one is a
wrong filing. And because the count is per-turn, the view degrades with use: the
projects where the agent is most active are the ones whose file lists become
least usable.

## Suggestion

Collapse the listing on `(created_in_scope, sha256)`: one row per document,
earliest `createdAt`, a count that expands to the history, and the set of
contributors. `FileListItem` (`src/api/app-types.ts:522`) already carries
`direction`, `ownerScopeId` and `createdAt`, so the surfaces need no new
endpoint — but the contributors line does require one addition:
`FileListItem` has no `createdBy` today, so either it gains that field or the
list route returns the grouped rows with authors included.

If a per-turn row is wanted as the underlying ledger — and it is a reasonable
thing to keep — then the fix is to stop treating that ledger as the file list.
Either way, adding `(created_in_scope, sha256)` alongside the existing
`(created_in_scope, created_at …)` partial index would make the grouped read
cheap without disturbing the current ordering.

I can test a branch against a live deployment; the reproduction above is two
people dropping the same file into a project and asking the agent to hand it
back.

yzxcj797:
Fixes #601.

## What this does

The artifact table is a **per-turn ledger** — `fileArtifactId` seeds from the run, so the same document re-attached by a later turn (or handed back by the agent) is a new row by construction, and `ON CONFLICT (id) DO NOTHING` only makes *within-run* retries idempotent. Nothing collapsed rows across turns, so a project's Files view filled with byte-identical, same-named, indistinguishable entries — one per turn that touched the document, degrading fastest exactly where the agent is most active (the issue's finance scope: three identically-named reimbursement schedules with no marker of which arrived from whom).

Per the issue's framing, the fix **keeps the ledger and stops treating it as the file list**:

- **`ListOwnedOptions.distinctByContent`** (opt-in): one row per document — the earliest registration of each `sha256` (the issue's "earliest `createdAt`" representative). Null-hash rows stay distinct via their id as the group key (`COALESCE(sha256, 'id:' || id)`), so the guard can never collapse unrelated rows.
  - Postgres: `DISTINCT ON (COALESCE(sha256, 'id:' || id)) … ORDER BY key, created_at ASC, id ASC` in a subquery, re-sorted `created_at DESC` outside — same ordering contract as today.
  - Memory store: the same grouping as a group-then-sort pass.
  - **Cursor correctness**: grouping happens *before* the cursor filter, so a document already emitted on a prior page cannot reappear through a later duplicate that happens to follow the cursor — the naive "filter then group" would straddle groups across pages.
- **`filesForViewer` passes it** — owned/shared and in-scope viewer views show one row per document. **Admin/ledger listings keep raw rows**: the admin files export and the per-user admin view are the audit surfaces where every registration should still be visible.
- **`FileListItem` gains `createdBy`** — who first registered the document, surfaced from the representative row. In the issue's words: three identical copies with no visible marker of which arrived from whom "is not a cosmetic problem — picking the wrong one is a wrong filing."

Not included (deliberately): the contributor-*set* aggregation and per-document occurrence count from the issue's fuller sketch — both need a second grouped query and UI to expand into; the row collapse plus `createdBy` removes the wrong-filing risk and the unbounded growth, and the rest builds cleanly on `distinctByContent` if wanted.

## Tests

`test/file-artifact-store.test.ts` (memory store, always runs): three same-bytes registrations across turns, owners, and directions (`in`/`in`/`out`, the issue's exact shape) collapse to the earliest row (`createdBy: sales`, `direction: in`, earliest `createdAt`); the raw ledger still returns all 4 rows; a second document stays distinct; cursor pagination over grouped pages never repeats a document across pages. **Fails on `main`** (no grouping exists).

`test/postgres-file-artifact-store.test.ts`: the same assertions against the `DISTINCT ON` path (skipped without `DATABASE_URL`, runs in CI).

Regression: the four file-surface suites (`files-http`, `reach-files`, `file-share`, `file-sharing`) pass 56/62 both with and without the change in this environment — the 6 failures are identical pre-existing Windows-only issues (fs `rename` EPERM); zero regressions. `tsc --noEmit` clean.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/617"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789788997&installation_model_id=19911&pr_number=617&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F617&signature=32804b5e32e4f97b303201be2f917eb0a974a7b4f28aec7a86c9dd64c2b0373b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/api/app-helpers.ts
+++ b/src/api/app-helpers.ts
@@ -318,8 +318,12 @@ export function createAppHelpers(deps: AppDeps, app: App) {
     inScope?: ScopeId,
   ): Promise<FileListPage> {
     const myScopes = await currentResourceScopesForViewer(principalId);
+    // Viewer-facing views show one row per document, not the per-turn
+    // ledger: an agent re-attaching the same file each turn must not grow
+    // the list without bound (#601). Admin/ledger listings keep raw rows.
     const owned = await deps.files.listOwnedByScopes(myScopes, {
       ...opts,
+      distinctByContent: true,
       ...(inScope ? { createdInScope: inScope } : {}),
     });
     const handles = await deps.acl.handlesFor(myScopes);
--- a/src/api/app-types.ts
+++ b/src/api/app-types.ts
@@ -574,6 +574,9 @@ export interface ContextSummary {
 interface FileListItem {
   id: string;
   ownerScopeId: ScopeId;
+  /** Who first registered this document. With content-grouped listings this
+   * is the author of the earliest row per document (#601). */
+  createdBy?: string;
   name: string;
   mimetype: string;
   sizeBytes: number;
@@ -624,6 +627,7 @@ export function toFileItem(a: FileArtifact): FileListItem {
   return {
     id: a.id,
     ownerScopeId: a.ownerScopeId,
+    ...(a.createdBy ? { createdBy: a.createdBy } : {}),
     name: a.name,
     mimetype: a.mimetype,
     sizeBytes: a.sizeBytes,
--- a/src/files/file-artifact-store.ts
+++ b/src/files/file-artifact-store.ts
@@ -54,6 +54,14 @@ export interface ListOwnedOptions {
   cursor?: string;
   includeDisabled?: boolean;
   createdInScope?: ScopeId;
+  /**
+   * Collapse per-turn duplicate rows: one row per document (the earliest
+   * registration of each sha256; null-sha rows stay distinct). The artifact
+   * table is a per-turn ledger — a document re-attached each turn is a new row
+   * by construction — so USER-FACING file views pass this to stop the file
+   * list growing with use. Ledger/admin listings leave it off (#601).
+   */
+  distinctByContent?: boolean;
 }
 
 export interface FileArtifactStore {
@@ -168,9 +176,25 @@ export function createMemoryFileArtifactStore(byteStore: DurableByteStore): File
       const set = new Set(scopes);
       const limit = clampLimit(opts?.limit);
       const cursor = opts?.cursor ? decodeCursor(opts.cursor) : null;
-      const all = [...rows.values()]
+      let all = [...rows.values()]
         .filter((r) => set.has(r.ownerScopeId) && (opts?.includeDisabled || r.enabled))
-        .filter((r) => opts?.createdInScope == null || r.createdInScope === opts.createdInScope)
+        .filter((r) => opts?.createdInScope == null || r.createdInScope === opts.createdInScope);
+      if (opts?.distinctByContent) {
+        // Group FIRST (over every matching row, cursor ignored), then apply
+        // the cursor to the representatives: a group whose earliest row was
+        // already emitted on a prior page must not reappear via a later
+        // duplicate that happens to follow the cursor (#601).
+        const earliest = new Map<string, FileArtifact>();
+        for (const r of all) {
+          const key = r.sha256 ?? `id:${r.id}`;
+          const cur = earliest.get(key);
+          if (!cur || r.createdAt < cur.createdAt || (r.createdAt === cur.createdAt && r.id < cur.id)) {
+            earliest.set(key, r);
+          }
+        }
+        all = [...earliest.values()];
+      }
+      all = all
         .filter((r) => (cursor ? afterCursor(r, cursor) : true))
         .sort((a, b) => b.createdAt - a.createdAt || idOrderDesc(a.id, b.id));
       const page = all.slice(0, limit);
--- a/src/files/postgres-file-artifact-store.ts
+++ b/src/files/postgres-file-artifact-store.ts
@@ -132,18 +132,39 @@ export function createPostgresFileArtifactStore(
         params.push(opts.createdInScope);
         filters.push(`created_in_scope = $${params.length}::text`);
       }
-      if (cursor) {
-        params.push(cursor.createdAt, cursor.id);
-        filters.push(`(created_at, id) < ($${params.length - 1}::bigint, $${params.length}::text)`);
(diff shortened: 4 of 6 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/774 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

Issue #604, opened 2026-08-19
admin error log stays empty through real failures


avaldez-gr:
I've been running a two week eval, solo, before pushing this out to others. Every real failure I hit left the admin error log completely empty. I know there's a family of open issues that echo this (#375 , #377 , #602 ). So maybe the useful framing is - the error logs intake is too narrow to catch the failures that actually happen.

These are the ones i'm facing:
- An app publish that crashed at boot (bad `entrypoint`). The deployment kept reporting "running" with a stale endpoint. #442 's preflight covers the entrypoint case now, but a crash at boot after a successful publish still shows up nowhere.
- The sandbox exec transport degraded after one full length (300s) foreground command. After that every blocking command over ~10s failed instantly with `fetch failed` while the container stayed healthy and fast commands worked fine. NOthing recorded. I see #469 added detection and auto-restart on the sprite backend, but on the local docker backend its still totally silent.
- A run that died mid task with `stopped:true` and an empty reply. The runs view shows it as done, attempts 1.

I get that the log can't record what the process didn't survive. But "failed inside a run that completed" and "failed above the run layer" both route to nothing right now. If deploy boot-crashes, exec-transport faults on all backends, and abnormal run termination each wrote an error event, the error page would be the first place I check instead of the last.

Issue #603, opened 2026-08-19
dev supervisor: if a child dies nothing restarts it and nothing records it, and there's no foreground mode for systemd


avaldez-gr:
Ran qm solo for a couple of weeks on a small cloud VM, started from a systemd unit that calls the dev CLI's up. The whole supervisor tree (core, portal, web) died on me twice in that window, and both times I only found out days later.

A few things stack up to make this painful:
- up always detaches the supervisor (`spawnDetached` in `scripts/dev/cli.ts`). There's no foreground mode, so the best you can do in systemd is `Type=oneshot` with `RemainAfterExit=yes`. When the tree died, the unit just sat there saying "active (exited)" and `Restart= ` never had anything to react to.
- The supervisor can restart children through its control endpoint (`POST /restart`) and on env changes, but a child that dies on its own stays dead, and nothing durable gets written about it. Both times I couldn't reconstruct the cause afterwards because there was nothing to look at. The admin error log was empty across the whole window.
- From the outside, a dead instance looks exactly like a quiet day. The second outage ran six days before I noticed, and the only reason I noticed was that my scheduled crons had been queueing silently the whole time (#382 fixes how that backlog drains, but not noticing the outage in the first place).

What I'd love, either one: 
1.  forground mode for up so systemd can own the lifecycle with `Restart=always`
2. The supervisor restarting dead children and writing a crash event somewhere the admin UI actually shows. Right now I've got a systemd timer curling the portal every couple of minutes and restarting the unit on any non 200. It works, but it feels like something the platform should do itself.

yzxcj797:
One correction to the premise that changes where the remaining work is: **a child that dies on its own IS restarted on current main** — `Child.handleExit` (`scripts/dev/supervisor/children.ts`, in place since 2026-08-14) schedules an automatic restart with exponential backoff (1s → 30s cap), parks a child after 5 consecutive crashes, and resets the crash counter after 60s of healthy runtime. That's independent of the `/restart` control endpoint and env-change restarts.

So for your two outages, the interesting question is which of these actually died:

1. **A child crashed and the supervisor restarted it** — then the restart half of your ask is already working, and what's missing is exactly the *recording* half: the crash line lands in `supervisor.log` only, nothing reaches the admin UI. That gap is shared with #604 (the admin error log's intake is too narrow to show real failures), and the durable crash-event path should be designed together with that issue's intake fix rather than bolted onto the dev supervisor alone.
[...]

PR #620, opened 2026-08-20
feat(dev): foreground mode for `up` so a service manager owns the lifecycle

yzxcj797:
Implements the foreground half of #603 (option 1 in the issue).

## What this does

`up` always detached the supervisor (`spawnDetached`), so the best a systemd unit could do was `Type=oneshot` + `RemainAfterExit=yes` — when the tree died on its own, the unit sat there "active (exited)" and `Restart=` had nothing to react to. `dev up --foreground` fixes that:

- the supervisor is spawned **attached** (new `spawnForeground` in `scripts/dev/lib/proc.ts` — the mirror of `spawnDetached`: stdio inherited so output lands on the caller's console or journald, child not unref'd, handle returned);
- after a successful boot the CLI **waits on the supervisor's exit** instead of returning;
- `SIGTERM`/`SIGINT` (a stop command, Ctrl-C) are forwarded to the supervisor so its own graceful tree teardown runs, and the CLI exits **0**;
- a supervisor that dies **on its own** exits non-zero (`EXIT.internal`), which `Restart=always` reacts to — the "active (exited)" silent-death failure mode this flag exists to remove.

```ini
[Service]
Type=simple
ExecStart=/path/to/qm dev up --foreground --org myorg
Restart=always
```

The detached mode is byte-for-byte unchanged (same `spawnDetached` call, same log file); the flag only swaps the spawn and adds the post-boot wait. Boot failure paths (supervisor never came up → killed, slot stolen, etc.) are untouched.

## Tests

`spawnForeground` unit tests in `test/dev-cli-lib.test.ts`: an attached child's exit code is observable through the returned handle (a `process.exit(3)` child resolves to 3 — the contract `Restart=always` depends on), and empty argv fails loud like `spawnDetached`. File: 19/19.

## Scope note

The crash-restart half of the issue — the supervisor itself respawning dead children and writing a durable crash event the admin UI shows — is a separate, larger change to the supervisor's lifecycle (it intersects with the ambient detection work and the admin error log's intake, #604) and is deliberately not bundled here; the foreground mode makes the OS-level supervisor the owner of exactly that failure instead.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/620"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789797155&installation_model_id=19911&pr_number=620&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F620&signature=f031ce7c973e7be23a31427477cc7af09b09ca9c25fbfe5cf7cf888b4562abe6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/scripts/dev/cli.ts
+++ b/scripts/dev/cli.ts
@@ -29,7 +29,7 @@ import {
   takenSummary,
 } from "./lib/lease.ts";
 import { callerEnvSnapshot, currentBranch, repoRoot } from "./lib/envctx.ts";
-import { killTree, pidAlive, portHolders, spawnDetached } from "./lib/proc.ts";
+import { killTree, pidAlive, portHolders, spawnDetached, spawnForeground } from "./lib/proc.ts";
 import {
   resolveSocketPath,
   streamBootEvents,
@@ -60,6 +60,7 @@ function parseCli() {
         sandbox: { type: "string", default: "auto" },
         "no-slack": { type: "boolean", default: false },
         "no-watch": { type: "boolean", default: false },
+        foreground: { type: "boolean", default: false },
         org: { type: "string" },
       },
     });
@@ -75,7 +76,7 @@ const command = positionals[0] ?? "up";
 const store = poolStore();
 
 const commandOptions: Record<string, readonly string[]> = {
-  up: ["json", "force", "strict", "rotate", "sandbox", "no-slack", "no-watch", "org"],
+  up: ["json", "force", "strict", "rotate", "sandbox", "no-slack", "no-watch", "org", "foreground"],
   down: ["json"],
   status: ["json"],
   restart: ["json"],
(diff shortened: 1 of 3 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/773 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #621, opened 2026-08-20
feat(mcp): negotiate a session for streamable HTTP servers

koreankop:
## Summary

The MCP HTTP client never establishes a session, so it cannot talk to servers built on the MCP SDK's `StreamableHTTPServerTransport`. Those servers answer every request with HTTP 400 `No valid session ID provided`, which surfaces here as `mcp tools/list failed (HTTP 400)` — the server is unusable, not degraded.

This change negotiates a session behind a single memoized promise:

- `initialize`, then `notifications/initialized`, resolving to the session id.
- Callers receive that id and pass it explicitly into each request instead of reading shared mutable state, so concurrent first callers share one session rather than racing to mint several (and orphaning the losers server-side).
- A request whose session is gone (HTTP 404) invalidates only the negotiation it actually used, so several in-flight callers converge on one replacement session and each retries once.

Deliberate boundaries:

- **Sessionless servers are untouched.** An `initialize` that fails or returns no session id resolves to `null`, the header is omitted, and the request shape matches the previous behavior exactly. A server that answers `initialize` with `Method not found` keeps working.
- **401/403 on initialize still throws**, so an auth failure is not silently read as "this server has no sessions".
- **Only 404 retries.** A 400 is the server's verdict on the request itself; re-sending it would deliver a non-idempotent `tools/call` twice.

## Test plan

`node --experimental-test-module-mocks --test test/mcp-connectors.test.ts` — 14/14 pass, `tsc --noEmit` and `eslint` clean. New cases:

- session negotiation: asserts the `initialize` → `notifications/initialized` → call order and that the negotiated id rides on later requests;
- sessionless server: asserts no `mcp-session-id` header is ever sent, across two calls;
- server without `initialize`: asserts tools still list;
- 401 on initialize: asserts it surfaces instead of falling back;
- three concurrent first calls: asserts exactly one `initialize`;
- session dropped under three concurrent callers: asserts all three recover on exactly one replacement session;
- genuine 400 on `tools/call`: asserts no retry and no re-send.

Also exercised against a live server: a stdio MCP server bridged to streamable HTTP, which returned 400 for every call before this change and now lists and calls tools normally.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/621"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789799721&installation_model_id=19911&pr_number=621&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F621&signature=ec2bd3a6b2af664e654f1db6046a80aa8ca76169479c89ad763cae15d6af4dde"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/mcp/mcp-client.ts
+++ b/src/mcp/mcp-client.ts
@@ -1,14 +1,3 @@
-// Generic Model Context Protocol (MCP) client — HTTP transport only.
-//
-// Speaks JSON-RPC 2.0 over a single POST endpoint, accepting both plain JSON
-// and SSE-framed responses (the two response shapes the spec's streamable
-// HTTP transport allows). Supports three auth modes: none, static bearer
-// token, and OAuth2 client-credentials minted against `<base>/token`.
-//
-// This is the transport layer only: no registry, no tool injection, no
-// policy. See mcp-tool-service.ts for the layer that turns registered
-// servers into agent tools.
-
 const TOKEN_SKEW_MS = 60_000;
 const MCP_ACCEPT = "application/json, text/event-stream";
 
@@ -131,6 +120,7 @@ export function createMcpClient(opts: {
   const host = hostOf(base);
   let cached: CachedToken | null = null;
   let rpcId = 0;
+  let session: Promise<string | null> | null = null;
 
   async function mintToken(clientId: string, clientSecret: string): Promise<string> {
     if (cached && now() < cached.expiresAt - TOKEN_SKEW_MS) return cached.accessToken;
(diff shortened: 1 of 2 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/772 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #622, opened 2026-08-20
fix(mcp): inline internal $refs in tool schemas at ingest

koreankop:
## Summary

A tool whose input schema refers to its own `$defs` or `properties` by JSON pointer breaks every turn once its schema is re-rooted into a model provider's tool list: the pointer still reads `#/properties/...`, but that path no longer exists at the new root. The provider rejects the entire tool list — Anthropic answers `Reference not found` — so turns fail even when they never touch the offending tool. Registration does not catch it: `tools/list` and the admin probe both succeed, because the re-rooting happens later.

This change inlines internal refs when a server's tools are ingested:

- Keywords beside a `$ref` still apply, so `{"$ref": "#/$defs/int", "minimum": 1, "maximum": 10}` keeps its bounds and description instead of widening to the bare target.
- External, unresolvable, and cyclic refs become a permissive `{}`; a `false` subschema stays restrictive as `{"not": {}}`.
- `$defs`/`definitions` are dropped at every level once inlined, not only at the root.
- Pointers are percent-decoded before `~1`/`~0` unescaping, per RFC 6901.

Expansion is bounded. Inlining duplicates a target at every use site, so an ordinary schema with a reused type — what `zod`'s `toJSONSchema({reused: "ref"})` and pydantic emit by default — can expand by orders of magnitude; a 2 KB schema of depth 8 with 8 fields per level expanded to hundreds of MB and stalled the event loop for tens of seconds, and deep plain nesting could exhaust the stack. A node budget and a depth cap now degrade those to a permissive subschema instead. Since `refresh()` runs on the main loop at construction, on registry change, and on a timer, that bound matters for availability.

Finally, the result is always an object schema: a root-level `$ref` that cannot be resolved falls back to a permissive object rather than an untyped `{}`, which would otherwise reach the provider as a tool with no object schema and reproduce the original failure by another route.

## Test plan

`node --experimental-test-module-mocks --test test/mcp-connectors.test.ts` — 16/16 pass, `tsc --noEmit` and `eslint` clean. New cases:

- ingest through the tool service: multi-hop chain, unresolvable ref, external ref, root `$defs` removal, no `$ref` anywhere in the result;
- keywords beside a `$ref` survive;
- always an object schema: root `$ref`, `#`, and a non-object schema all fall back; a typeless schema with `properties` gains `type: "object"`;
- bounded expansion: depth 8 × width 8 stays under 2 MB and 2 s;
- 5000-deep plain nesting does not exhaust the stack;
- escaped (`~0`, `~1`, `%20`) and array-index pointers resolve;
- direct and mutual cycles terminate;
- nested `$defs`/`definitions` dropped, and a `__proto__` key neither reparents the output nor survives;
- a `false` subschema stays restrictive.

Observed in practice with a public MCP server whose schemas carry `#/properties/...` refs: with it enabled every turn failed; with this change the same server's tools ingest and run.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/622"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789799729&installation_model_id=19911&pr_number=622&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F622&signature=221b6c88e3ddfc801b0b055e4fdc695d0b0a00835d41565c5da6b7ab34633fcd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/mcp/mcp-tool-service.ts
+++ b/src/mcp/mcp-tool-service.ts
@@ -14,6 +14,8 @@ import type { McpServer, McpServerStore } from "./mcp-server-store.ts";
 const REFRESH_INTERVAL_MS = 5 * 60_000;
 const MAX_TOOLS_PER_SERVER = 64;
 const MAX_RESULT_CHARS = 60_000;
+const MAX_SCHEMA_NODES = 20_000;
+const MAX_SCHEMA_DEPTH = 100;
 
 export interface McpToolDescriptor {
   /** Namespaced tool name exposed to the model, e.g. "salesforce_query". */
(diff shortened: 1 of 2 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/771 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #623, opened 2026-08-20
feat(orchestrator): record surface-tool replies as 'delivered', not 'silent'

### LINKED ISSUE #609: A turn's outcome isn't recorded faithfully: silent means two opposite things, and suppressed turns are absent from turn_metrics
# A turn's outcome isn't recorded faithfully: `silent` means two opposite things, and suppressed turns are absent from `turn_metrics` entirely

I built a harness task against the Slack ambient path and it produced a wrong
verdict three times running, each time because the durable record does not say
what happened to a turn. The bugs were mine and are fixed, but each one was
downstream of the same thing: an operator, a dashboard, or an alert reading
qm's own tables cannot distinguish "the agent answered" from "the agent was
suppressed."

Numbers below are from my deployment, ~1,274 turn_metrics rows
(refreshed 2026-09-02; the counts in the original filing were smaller and the
ratios unchanged).

## 1. `status: "silent"` covers both suppression and success

`src/core/orchestrator.ts:1581` — the ambient reply gate declined, nothing
happened:

```ts
return reactions
  ? { status: "react", sessionId: session.id, reactions }
  : { status: "silent", sessionId: session.id };
```

`src/core/orchestrator.ts:2927` — the turn completed normally and delivered its
reply through a surface tool, which is how every ordinary Slack answer comes
back:

```ts
} else if (input.surfaceTools && surfaceToolDeps && !strictReadOnly) {
  finalResult = { status: "silent", sessionId: session.id, ...(result.stopped ? { stopped: true } : {}) };
}
```

There is a third producer at `:2926` — a scheduled (poll-fired) turn where the
model itself chose to say nothing:

```ts
} else if (isPollFire && !outbound.attachments.length && isSilentPollReply(reply)) {
  finalResult = { status: "silent", sessionId: session.id };
}
```

That one is at least close to genuine silence, but it lands in the record as
the same value, so it cannot be separated from suppression either.

Same value, opposite outcomes. On this deployment:

```
runs.result status   ok 496 | silent 106 | refused 15 | failed 14 | react 3
```

Those 106 are unresolvable from the record. Some are agents that answered a
colleague; some are messages qm decided not to answer at all. Nothing stored
separates them.

I hit this concretely: a control run where qm replied correctly in-thread came
back `silent`, and scoring on that field marked a working exchange as a failure.
Reading the Slack transcript instead has the mirror problem — it cannot tell a
declined turn from a model that ran and chose to say nothing. Neither source
answers the question on its own.

## 2. A declined turn never reaches the metrics sink

The gate returns at `:1565`. The metrics emit is at `:2736`, ~1,200 lines later
in the same function. So a turn the gate declines writes:

- no `turn_metrics` row
- no `session_llm_requests` row (the detection call is passed only
  `recordModelCall`, which lands in the in-memory ring in
  `src/model/model-gateway.ts` and is never persisted)

~~Its sole durable trace is `runs.result`.~~ *(Corrected 2026-09-02 — see the
comment below. `audit_log` also records the decision as `turn.silent` /
`turn.react`. That second record is what let me verify this section at record
level rather than by inference, and the two sources agree exactly.)*

```
turn_metrics status   capture 637 | ok 637
```

No `react`, no `silent`, no `refused`, no `failed` — in 1,274 rows. And it is not
only the gate's declines that are missing: `refused` (15) and `failed` (14) have
no `turn_metrics` row either. Three outcome classes never reach this table.
Which brings me to:

## 3. `turn_metrics.status` is hardcoded, so even the recorded turns don't carry their outcome

```ts
// src/core/orchestrator.ts:2740
status: pausing ? "paused" : "ok",
```

`finalResult` is computed a hundred lines below and never reaches the sink. The
arithmetic confirms it exactly: 506 `ok` + 131 `silent` in `runs` = the 637 `ok`
in `turn_metrics`. Turns that returned `silent` are recorded as `ok`.

So the natural table for an operational dashboard — the one with `total_ms`,
`ttft_ms`, `model_calls` and the rest — reports every turn it knows about as
successful, and doesn't know about the suppressed ones at all.

## 4. `detect_ms` exists but can only ever record a "yes"

```
turn_metrics.detect_ms populated:  4 / 1274
```

The column is declared in `src/admin/postgres-metrics-sink.ts:31` and assigned
at `:2736`. Because the declining path returns at `:1565`, `detect_ms` is
written only when the gate said respond — the case least in need of
investigation. Every declined decision, and its latency, is dropped.

## Why this is worth fixing beyond tidiness

The ambient gate is a model call that decides whether a person gets an answer.
On this deployment it silently dropped an answer to a question qm itself had
asked and promised to report on. There is currently no query that would surface
that class of event, no metric that would trend it, and nothing that records
which model made the call — `judgeModelId` resolves to
`auxiliaryModelFor(orgBaseModelId())`, so for an Anthropic org model the decision
runs on `claude-haiku-4-5`, but that is derived from source, not observed.

An operator who suspects qm is ignoring people has, today, only `runs.result`
and a `status` that means two opposite things.

## Suggestions

1. Split the two `silent` cases. `delivered` (or reuse `ok`) for a surface-tool
   delivery, and keep `silent` for suppression — or add a `reason` alongside.
   One value for both makes every downstream consumer wrong by default.
2. Emit a `turn_metrics` row on the early returns, so declined turns exist in
   the same table as the rest, with `detect_ms` and the gate's verdict.
3. Carry `finalResult.status` into the sink rather than hardcoding `"ok"`.
4. Persist the detection call's model on `session_llm_requests` like any other
   model call, so which model made the decision is answerable after the fact.

(1) and (3) look small and would do most of the work.

Happy to test a branch; I have a live deployment and a harness that drives this
path — [`ambient-reply.mjs`](https://github.com/ianTPE/qm-model-eval), which
also documents the three ways this record misled me while I was building it.
Related: #607 (the gate's decision itself) and #608 (the same hand-off seen from
another session).

yzxcj797:
Fixes #609 — suggestions (1) and (3) from the issue, which the reporter identified as "small and would do most of the work".

## What changes

**`TurnResult` gains `"delivered"`.** The surface-tool branch — the ordinary shape of a Slack answer, where the reply goes out through the `post` tool rather than the TurnResult reply field — previously recorded `status: "silent"`, the *same value* the ambient gate writes when it declines to answer. A delivered answer and a suppressed turn were indistinguishable in `runs.result`; the reporter's harness scored a correct in-thread reply as a failure three times running because of it. Now:

| producer | status |
| --- | --- |
| ambient gate declined | `silent` (unchanged — means exactly this) |
| poll fire, model said nothing | `silent` (unchanged) |
| surface-tool delivery | **`delivered`** (was `silent`) |
| ordinary reply | `ok` (unchanged) |

**The metrics sink stops hardcoding `"ok"`.** `turn_metrics.status` now mirrors the finalResult branches (same primitives, same order — including collect-mode's `ok`-with-approvals, pinned by the existing test): delivered turns land as `delivered`, silent polls as `silent`, approval pauses as `paused`/`pending_approval`. Previously ALL of them landed as `"ok"` (the reporter's 1,200-turn dump: `capture 600 | ok 600`, nothing else).

**Consumers taught the new value** (audited: every `status === "silent"` / `!== "ok"` site):
- Slack turn handler: `delivered` settles the ack and stands down — the reply already posted, delivering again would double-post (exactly the pre-split silent behavior).
- Trigger runner: `delivered` counts as a success in failure classification, and stands down on forwarding (the surface tool already posted at the destination).

Suggestions (2) and (4) — metrics rows for gate-declined early returns, persisting the detection call's model request — are larger changes to the early-return path and are deliberately not included.

## Tests

- `a surface-tool turn records status 'delivered', not 'silent'` — **fails on `main`** (`silent`).
- `the metrics sink carries the turn's outcome instead of hardcoding 'ok'` — the `turn_metrics` row for a surface-tool turn reads `delivered`. **Fails on `main`** (`ok`).
- `a gate-declined ambient turn still records 'silent'` — pins that the split doesn't blur suppression.

Full `orchestrator.test.ts`: 69 passed / 60 pre-existing environment failures, identical set on clean `main` (where the two new tests fail); trigger suites 39/39; `tsc --noEmit` clean.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/623"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789805038&installation_model_id=19911&pr_number=623&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F623&signature=2e884f1014dd5310a48ad973eaabdd048f81e8234ea02ad22572318610017ba3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -2762,7 +2762,24 @@ export function createOrchestrator(deps: OrchestratorDeps): Orchestrator {
           ...(compactMs !== undefined ? { compactMs } : {}),
           ...(typeof input.queueMs === "number" ? { queueMs: Math.max(0, input.queueMs) } : {}),
           ...(resumedFromSeq !== undefined ? { resumedFromSeq } : {}),
-          status: pausing ? "paused" : "ok",
+          // Mirror the finalResult branches below (same primitives, same
+          // order) so the metrics row says what actually happened instead of
+          // hardcoding "ok" — delivered answers, silent polls, and approval
+          // pauses all used to land here as "ok" (#609).
+          status:
+            pausing
+              ? "paused"
+              : isPollFire && result.silent && result.pausedOnApproval !== true
+                ? "silent"
+                : result.pendingApprovals?.length
+                  ? turnCompleted
+                    ? "ok"
+                    : "pending_approval"
+                  : isPollFire && !outbound.attachments.length && isSilentPollReply(reply)
+                    ? "silent"
+                    : input.surfaceTools && surfaceToolDeps && !strictReadOnly
+                      ? "delivered"
+                      : "ok",
           scopeLabel: scopeId,
           provisioned:
             !!box.handle ||
@@ -2950,7 +2967,15 @@ export function createOrchestrator(deps: OrchestratorDeps): Orchestrator {
         } else if (isPollFire && !outbound.attachments.length && isSilentPollReply(reply)) {
           finalResult = { status: "silent", sessionId: session.id };
         } else if (input.surfaceTools && surfaceToolDeps && !strictReadOnly) {
-          finalResult = { status: "silent", sessionId: session.id, ...(result.stopped ? { stopped: true } : {}) };
+          // The reply was delivered through a surface tool — the ordinary
+          // shape of a Slack answer. Recording it as "silent" made a
+          // delivered answer indistinguishable from a suppressed turn in
+          // runs.result (#609).
+          finalResult = {
+            status: "delivered",
+            sessionId: session.id,
+            ...(result.stopped ? { stopped: true } : {}),
+          };
         } else {
           finalResult = {
             status: "ok",
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -2766,20 +2766,19 @@ export function createOrchestrator(deps: OrchestratorDeps): Orchestrator {
           // order) so the metrics row says what actually happened instead of
           // hardcoding "ok" — delivered answers, silent polls, and approval
           // pauses all used to land here as "ok" (#609).
-          status:
-            pausing
-              ? "paused"
-              : isPollFire && result.silent && result.pausedOnApproval !== true
-                ? "silent"
-                : result.pendingApprovals?.length
-                  ? turnCompleted
-                    ? "ok"
-                    : "pending_approval"
-                  : isPollFire && !outbound.attachments.length && isSilentPollReply(reply)
-                    ? "silent"
-                    : input.surfaceTools && surfaceToolDeps && !strictReadOnly
-                      ? "delivered"
-                      : "ok",
+          status: pausing
+            ? "paused"
+            : isPollFire && result.silent && result.pausedOnApproval !== true
+              ? "silent"
+              : result.pendingApprovals?.length
+                ? turnCompleted
+                  ? "ok"
+                  : "pending_approval"
+                : isPollFire && !outbound.attachments.length && isSilentPollReply(reply)
+                  ? "silent"
+                  : input.surfaceTools && surfaceToolDeps && !strictReadOnly
+                    ? "delivered"
+                    : "ok",
           scopeLabel: scopeId,
           provisioned:
             !!box.handle ||
--- a/src/slack/turn-handler.ts
+++ b/src/slack/turn-handler.ts
@@ -513,6 +513,14 @@ export function createTurnHandler(deps: {
       return;
     }
 
+    if (result.status === "delivered") {
+      // The surface tool already posted the reply (the ordinary Slack
+      // answer shape); delivering it here too would double-post. Settle the
+      // ack and stand down, same as the pre-split silent handling (#609).
+      await settleAck();
+      return;
+    }
+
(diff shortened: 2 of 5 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/770 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

Issue #602, opened 2026-08-19
When a model provider fails: no retry on the pi harness, no cron backoff, no failover on transient errors, and /healthz stays green


ianTPE:
# When a model provider fails: no retry on the pi harness, no cron backoff, no failover on transient errors, and /healthz stays green

I went looking for what a qm deployment does when its model provider has a bad
ten minutes — not when it's misconfigured, but when a working deployment's
provider starts returning 429s or 5xx. The answer is worse than I expected, and
the pieces compound. Splitting them into sections since the fixes are
independent, but the reason to read them together is that they combine into one
outage story.

Environment: self-hosted, Postgres-backed, org on the `claude` harness, other
scopes pinned to `pi` models across Anthropic, OpenAI, DeepSeek, Moonshot, Z.ai
and DashScope.

## 1. The `pi` harness treats every provider error as terminal

```ts
// src/harness/pi-harness.ts:838
export function piLastAssistantTextOrThrow(session: AssistantTextSession): string | undefined {
  const err = piAssistantError(session);
  if (err) throw new NonRetryableTurnError(err);
```

```ts
// src/harness/pi-harness.ts:844
export function piTurnError(session: AssistantTextSession, thrown: unknown, messagesBefore?: number): Error {
  const detailed = piAssistantError(fresh);
  if (detailed) return new NonRetryableTurnError(detailed);
```

`piAssistantError` (:831) fires on any assistant message with
`stopReason === "error"`, and `formatPiAssistantError` (:809) happily renders
the provider's own error type into the message. So `rate_limit_error` and
`authentication_error` take the same path — both become
`NonRetryableTurnError`, and `src/runs/worker.ts:76`
(`retry: !(err instanceof NonRetryableTurnError)`) parks the run without a second
attempt.

`max_attempts` defaults to 3 (`src/runs/postgres-run-store.ts:52`). On this
deployment it has almost never been reached:

```
attempts | error_attempts | max_attempts | runs
       1 |              1 |            3 |   13
       3 |              3 |            3 |    1
```

The single run that used its retries is the one `claude`-harness failure
("Claude Code returned an error result"); that harness only marks the wall-clock
timeout non-retryable (`src/harness/claude-harness.ts:708`) and lets everything
else retry. Every `pi` failure got exactly one attempt, including this one:

```
Model provider API error (rate_limit_error): [1113][Insufficient…]
```

A rate limit is the textbook retryable error and it killed the turn outright.
[...]

PR #624, opened 2026-08-20
feat(runs): retry transient provider errors with backoff

yzxcj797:
Fixes the pi-retry and cron-adjacent backoff sections of #602 (sections 1 and the retry-cadence half of 2). Failover (3) and /healthz (4) are larger design decisions, deliberately not included.

## Section 1: every pi provider error was terminal

`piAssistantError` fires on any assistant message with `stopReason === "error"`, and both `piLastAssistantTextOrThrow` and `piTurnError` wrapped the formatted message in `NonRetryableTurnError` — so `rate_limit_error` and `authentication_error` took the same path, and the runs worker (`retry: !(err instanceof NonRetryableTurnError)`) parked the run after one attempt. Your deployment table says it exactly: every `pi` failure got 1 attempt; the only run that ever reached 3 was the `claude`-harness one, whose harness already treats everything but wall-clock timeout as retryable.

**Fix:** classification by provider error type. `rate_limit_error`, `overloaded_error`, `timeout_error`, `api_error`, `server_error` (plus the less-common siblings: `internal_server_error`, `service_unavailable`, `temporarily_unavailable`, `model_service_overloaded`) become plain `Error`s the worker retries; `authentication_error` / `permission_error` / `invalid_request_error` and non-JSON errors stay `NonRetryableTurnError` — a second attempt would fail identically. The formatted message is unchanged in both classes (`Model provider API error (rate_limit_error): …` still greps the same); only the class carries the decision.

## The retry cadence: retries were immediate

Requeue sets `status='pending'` and the worker polls at 50ms — so even with retries, `maxAttempts=3` hits a rate-limited provider back-to-back, "feeding the thing that caused it" (your section 2 observation, applied to the run path).

**Fix:** an ERROR-driven requeue now carries a not-before — `retryBackoffMs(errorAttempts)` = 5s × 2ⁿ capped at 60s — stored in the pending row's `lease_expires_at` and honored by both claim paths (`claim` and `claimById`, memory + Postgres; nothing else sets `lease_expires_at` on pending rows today, and the reaper only touches `running` rows, so the overload is backwards-compatible). Lease-expiry requeues — your suspected-crash path — keep their immediate retry and their own poison-pill budget, unchanged.

## Tests (`pi-retryable-errors.test.ts`)

- rate_limit_error → **not** `NonRetryableTurnError` (the exact failure from your runs table), message still matches `/rate_limit_error/`.
- overload / api_error / server_error / timeout → retryable; authentication / permission / invalid_request → non-retryable; plain-text errors keep the historical class.
- `retryBackoffMs`: 5s → 10s → 20s … capped 60s.
- Memory-store behavior: an error-requeued run is **not claimable** inside the window (both `claim` and `claimById`); a fresh run claims immediately.

Existing suites updated for the intentional cadence change: `run-store.test.ts` / `worker-reaper.test.ts` pass `retryBackoffMs: () => 0` where they model immediate claim-after-fail (the knob the memory factory now exposes); the lease-expiry reaper tests needed nothing. Full set: 87/87, `tsc --noEmit` clean.

## What's deliberately not here

- **Failover** (section 3): `resolveRuntimeChoice` is a pure config read; making it provider-health-aware needs a health signal and a policy ("use B when A fails" needs somewhere to express B). Your note that the refusal-fallback machinery is proven and merely unwired is the right observation — wiring it to availability failures is a design PR of its own.
- **Cron tick backoff** (section 2's batch shape): one failing cron aborting its tick's batch deserves its own fix (isolation + per-cron backoff), separate from the run-level backoff landed here.
- **/healthz** (section 4): surfacing model reachability in a health endpoint is an ops contract decision (what does `ok` mean to your uptime monitor).

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/624"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789806938&installation_model_id=19911&pr_number=624&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F624&signature=57ed211cc7720e848f861af67c9a0f7b2953bf929e7732646366fdad83ffa430"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/harness/pi-harness.ts
+++ b/src/harness/pi-harness.ts
@@ -863,9 +863,61 @@ function piAssistantError(session: AssistantTextSession): string | null {
   return formatPiAssistantError(lastAssistant.errorMessage);
 }
 
+/**
+ * Provider error types that a retry can plausibly survive — the provider is
+ * alive but refusing (rate limit), degraded (overload / 5xx), or slow
+ * (timeout). Everything else (authentication, permission, invalid request…)
+ * fails identically on the next attempt, so those keep the non-retryable
+ * classification (#602).
+ */
+const RETRYABLE_PROVIDER_ERROR_TYPES = new Set([
+  "rate_limit_error",
+  "overloaded_error",
+  "timeout_error",
+  "api_error",
+  "server_error",
+  "internal_server_error",
+  "service_unavailable",
+  "temporarily_unavailable",
+  "model_service_overloaded",
+]);
+
+function piProviderErrorType(raw: string | undefined): string | null {
+  const message = raw?.trim();
+  if (!message) return null;
+  const jsonAt = message.indexOf("{");
+  if (jsonAt < 0) return null;
+  try {
+    const parsed = JSON.parse(message.slice(jsonAt)) as { error?: { type?: unknown } };
+    const providerType = typeof parsed.error?.type === "string" ? parsed.error.type.trim() : "";
+    return providerType || null;
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Classify a provider stop-error for retry purposes: transient provider
+ * failures (rate limit, overload, 5xx, timeout) become plain errors so the
+ * runs worker retries them with backoff; permanent ones (authentication,
+ * permission, bad request) stay non-retryable — a second attempt would fail
+ * identically (#602).
+ */
+function piTurnErrorFor(raw: string | undefined): Error {
+  const message = formatPiAssistantError(raw);
+  const providerType = piProviderErrorType(raw);
+  if (providerType !== null && RETRYABLE_PROVIDER_ERROR_TYPES.has(providerType)) {
+    // Same formatted message as the non-retryable path — the class carries
+    // the retry decision, the copy stays what operators grep for.
+    return new Error(message);
+  }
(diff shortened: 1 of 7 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/769 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #625, opened 2026-08-20
feat(cron): back a failing cron off instead of re-firing every tick

yzxcj797:
Fixes the cron-backoff section of #602 (the remaining half of section 2).

## What this fixes

A cron whose fire failed was retried by the schedule alone: nothing updated `lastFiredAt` on failure, so the next scheduler sweep (1s cadence) re-derived the same `scheduledAt` and fired again — your "every 1-5s until it recovers" observation, which during a rate-limit outage feeds the limit causing it. (The tick's per-fire isolation your issue also describes is already in place on main — `fireDue` catches per cron, so one failing cron does not abort its batch; that part is verified unchanged.)

## The change

Each cron now carries a failure backoff:

- a failed fire increments `failedAttempts` and sets `retryNotBefore = now + cronFailureBackoffMs(n)` — **30s × 2ⁿ, capped at 5 minutes** (a recurring cron probes a dead provider at most every few minutes, vs every tick);
- `due()` holds a cron out while `retryNotBefore` is in the future, whatever the schedule derives — the retry pacing beats the schedule during an outage;
- a **successful fire clears both fields** (`markFired`), so a recovered cron returns to its exact schedule immediately — no lingering penalty;
- the fire-log dedup you noted (keyed by fireKey, overwritten not appended) is untouched, so storage stays flat.

## Tests

- `a failing cron backs off exponentially instead of re-firing every tick` — first failure holds the cron quiet for 30s across four in-window ticks; the window passing re-attempts; the second failure doubles to 60s. **Fails on `main`** (no backoff exists).
- `a cron's backoff clears after a successful fire` — a flaky cron fails once, succeeds after the window, and carries no stale not-before.
- Updated: the starvation test (`a failing interval cron does not starve later due crons across ticks`) — its subject (the succeeding cron firing on schedule beside a failing neighbor) is unchanged and still asserted; the failing neighbor's sub-30s retry its expectation also pinned was precisely the outage-feeding behavior this PR removes, so the expectation now shows the neighbor held off.

Full cron suites: scheduler 53/53, store/queue/scope-shared/guardrails 55 pass + 2 pre-existing skips (0 fail), identical on clean `main` apart from the updated expectation; `tsc --noEmit` clean.

## Where this leaves #602

- ✅ section 1 (pi retryable classification) + run-level error-retry backoff — #624
- ✅ this PR (cron failure backoff)
- ⬜ section 3 (failover) and section 4 (/healthz) — design decisions, outlined in #624's description.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/625"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789807583&installation_model_id=19911&pr_number=625&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F625&signature=2b36785e1ce929cf7a52b499f12ac670e195d5332ae8d5fbf14c5fcb67cfd745"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/cron/cron-store.ts
+++ b/src/cron/cron-store.ts
@@ -46,9 +46,19 @@ export interface CronStore {
   recordFire(id: string, entry: CronFireLogEntry): Promise<void>;
   markFired(id: string, at: number, scheduledAt?: number): Promise<void>;
   markAttempted(id: string, at: number): Promise<void>;
+  markFailedAttempt(id: string, at: number): Promise<void>;
   claimSlot(id: string, scheduledAt: number, at: number): Promise<boolean>;
   unclaimSlot(id: string, scheduledAt: number, at: number, priorLastFiredAt: number | undefined): Promise<void>;
   due(now: number): Promise<Array<Cron & { scheduledAt: number }>>;
+}
+
+/**
+ * Per-cron failure backoff (#602): 30s exponential, capped at 5 minutes — a
+ * recurring cron with a failing provider is probed at most every few minutes
+ * during an outage instead of every scheduler tick.
+ */
+export function cronFailureBackoffMs(failedAttempts: number): number {
+  return Math.min(30_000 * 2 ** Math.max(0, failedAttempts - 1), 300_000);
 }
 
 function normalizeTitle(title: string | undefined): string | undefined {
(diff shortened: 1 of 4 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/768 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #628, opened 2026-08-20
feat(slack): refuse unresolved own-team members as their own state

yzxcj797:
Fixes defects 1 and 2 of #626 (the classification collapse and the invisible refusal); the `ok:false` propagation (1b) and event-driven refresh (3) are separate follow-ups per the issue's fix ladder.

## Defect 1 — the third state

`classifyUser` folded three states into two: confirmed-external → guest; confirmed-internal-resolved → allowed; and **own-team-but-email-invisible → also "guest"**. An org owner with a just-set email masqueraded as an external guest for the directory-propagation window, with every log line and the user-facing copy asserting "external".

Now `ActorAssertion` carries `principalUnresolved: true` for that third case:

- **Native flags still decide `isExternalGuest`** — a restricted/other-workspace/stranger member with a *visible* email remains an email-keyed external guest exactly as before (pinned by the pre-existing tests, now joined by a "native evidence wins" case). The third state never masks the flags.
- **Email keying is unchanged for everyone with a visible email.**
- **slack-id mode never sets it** (no email resolution exists there).

## Defect 2 — the refusal is its own state, and observable

The turn-handler gate refuses `principalUnresolved` **ahead of** the external-guest gate — equally fail-closed (never a fallback Slack-ID principal, which would mint a second internal identity with wrong scope/grants, exactly what email mode exists to prevent), but:

- the sender sees *"I couldn't verify your team membership (your email isn't visible to me). This usually resolves within a few minutes of the directory refreshing; try again shortly"* — naming the actual condition and its actual remedy, instead of "this conversation isn't fully internal", which sent the owner looking for a channel-setting problem;
- **every refusal logs on the success path** — the new `principalUnresolved` refusal warns with user/channel/timestamp, and the external-guest refusal (which previously left no trace unless `postEphemeral` *threw*) now warns too. In the incident, H1 (path never ran) / H2 (ephemeral posted but not displayed) / H3 (hung await before the ephemeral) were indistinguishable precisely because a successful refusal was invisible; the success-side log line makes the next one decidable.

## Tests

- Updated: "email mode fails closed to guest when a member has no visible email" → asserts `isExternalGuest: false` + `principalUnresolved: true`. **Fails on `main`** (guest).
- New: "native external evidence still wins over unresolved" — restricted + email-less reads as external guest, `principalUnresolved` unset. Pins that the third state cannot mask the flags.
- New: "slack-id mode never marks principalUnresolved".

[...]

Issue #626, opened 2026-08-20
Email identity mode refuses an unresolved own-team member as an external guest, and the refusal can be fully silent


ianTPE:
In email identity mode, an own-team member whose email the bot can't currently see is classified `guest`, and their direct @mention is refused. That refusal is *supposed* to send the sender an ephemeral "I can't respond here"; in the incident below it did not — the member (an org **owner**) got no turn, no reply, and no ephemeral at all. Two problems: identity uncertainty is disposed as "external guest" rather than "own-team, principal-unresolved"; and the one user-visible signal that should exist for the refusal never appeared, in a way the logs can't explain.

Line refs are against local HEAD (`f9aec5e`).

## The incident (evidence)

Self-hosted, Slack socket mode, email identity mode (`SLACK_IDENTITY_EMAIL=1`).

- `founder` (own-team, **owner**, `is_restricted=false`, email `founder@authikey.com`) sent `@qm …` twice (`mentions_self=t`). Both landed in `channel_messages` as `handled=f`, produced no turn and no `error_events`, and — user-confirmed — produced **no ephemeral and no reply of any kind**.
- The email had just been set and wasn't yet visible in the directory, so in email mode the member classified `guest` (~39-min window, dominated by Slack-side propagation of the new email).
- The only log line at each message's timestamp is the directory-snapshot warning (`src/slack/directory.ts:139`). No external-gate log, no `postEphemeral` attempt or failure, no turn.
- After the email became visible and the snapshot refreshed, the member classified `internal` (`directory_members.type`) and a later mention ran normally. The two earlier messages stayed `handled=f` — no backfill, permanently unanswered.

## Defect 1 — identity uncertainty is disposed as "external guest"

`classifyUser` first derives guest from Slack's own flags — missing record, `deleted`, `is_restricted`, `is_ultra_restricted`, `is_stranger`, team mismatch (`src/slack/identity.ts:46`). Then, in email mode, it *adds*:

```ts
// src/slack/identity.ts:59
else isGuest = true;   // email mode, no visible "@" email
```

So a confirmed own-team, unrestricted **owner** becomes `guest` purely because their email isn't currently visible — even though the plugin's own snapshot warning already calls this out ("…fail closed to guest", `directory.ts:139`). Email visibility is a workspace/plan-controlled policy, not evidence of external restriction; the native flags already carry that.

[...]

PR #629, opened 2026-08-20
fix(slack): preserve users.info failures instead of laundering them

yzxcj797:
Fixes the `ok:false`-propagation half of #626 (defect 1b). **Stacked on #628** (the third-state commit — `principalUnresolved` and the refusal gates come from there); merge that first.

## The laundering

A transient `users.info` failure returned `{ actor: { isExternalGuest: true }, ok: false }` — the flag existed, but both consumers discarded it:

- `classifyActor` returned only `.actor`, so the assertion arrived as a confident "external guest" with no trace of the failure;
- the direct-intake path (`turn-handler`) read `classifyUserCached(...)` and ignored `.ok` identically.

A single failed lookup therefore refused a member outright — in the incident, an org owner, with no signal that the classification was a placeholder rather than a verdict.

## The fix

- **The failure arm stops asserting guest**: `classifyUserCached`'s catch returns a `lookupFailed` actor (id-based placeholder, named for what it is). Distinguishable from every checked state — guest, internal, and (from #628) principal-unresolved.
- **`classifyActor` returns `ActorAssertion & { ok }`** — the flag rides along; consumers decide what a failed lookup means. The roster path already consumed `ok` (`resolveChannelMembership` sets `complete=false`); the two discard sites no longer can.
- **The direct-intake path acts on it**: fall back to `lastKnownClassification(userId)` — the LRU entry, which failures never populate, so the fallback is always a *prior success*, time-bounded by the cache TTL — logging the fallback; with no prior state, refuse as `principalUnresolved` (the third-state gate: fail closed, visibly, never a laundered "external").

## Tests

- `a failed users.info asserts lookupFailed, not external guest` — drives a failing client through `createDirectory`: `ok:false`, `lookupFailed:true`, `isExternalGuest` unset. **Fails on the base** (guest).
- `classifyActor preserves ok instead of discarding it` — the interface contract with a working client: `ok:true` rides on the returned assertion. **Fails on the base** (no `ok` on the return).

`slack-identity` 29/29 on this stack; the four identity suites 80/80 both with and without the patch's base state. `tsc --noEmit` clean.

## Remaining from #626

Defect 3 (event-driven refresh: `user_change`/`team_join` subscriptions + cache eviction) — needs the manifest change and the app-reinstall note the issue describes.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/629"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789826625&installation_model_id=19911&pr_number=629&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F629&signature=4fc5aa1d6c1cf232f6672fdd882748131efcf37c4ab3b7750902037cd2f46e21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/slack/directory.ts
+++ b/src/slack/directory.ts
@@ -65,7 +65,11 @@ export interface Directory {
   getUserSnapshot(client: any): Promise<{ byId: Map<string, CachedUser>; fetchedAt: number } | undefined>;
   forceDirectorySync(client: any, invalidateChannelId?: string, invalidatePrincipalId?: string): Promise<void>;
   classifyUserCached(client: any, userId: string): Promise<CachedUser & { ok: boolean }>;
-  classifyActor(client: any, userId: string): Promise<ActorAssertion>;
+  lastKnownClassification(userId: string): ActorAssertion | undefined;
+  classifyActor(
+    client: any,
+    userId: string,
+  ): Promise<ActorAssertion & { ok: boolean }>;
   getChannelInfo(client: any, channel: string): Promise<ChannelMeta | undefined>;
   channelMembership(
     client: any,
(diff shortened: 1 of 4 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/766 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

Issue #550, opened 2026-08-16
Keychain credentials cannot be materialized for a non-agent session (SSH or browser IDE) on a shared-scope sandbox


profd2004:
## The gap

The keychain models per-user credentials well: `keychain_credentials` (owner, `envKey`, `service`, `host`, `expiresAt`, `refresh`, encrypted secret), `keychain_grants` (credential → `audienceScopeId`, `once`/`standing`, active/used), `keychain_asks`, and `credential_usage` for audit. Materialization is deliberately ephemeral: `src/credentials/resident-paths.ts` defines `EPHEMERAL_CRED_DIR = "/tmp/agent-creds"` and symlinks `.netrc`, `.config/gh`, `.config/glab`, `.config/gcloud`, `.aws` into it, leaving `DURABLE_CREDENTIAL_PATHS = [".ssh", ".git-credentials"]` resident. Credentials exist in a sandbox only while a grant is materialized. That is the right shape.

The problem is that materialization is only reachable from the agent/web surface:

- `secret_drops.destination` is a `Destination` (`src/types.ts`), and the destinations that can be attached are the ones carried on a capability token minted for a turn (`src/api/capability-destination.ts`) — in practice `{"type":"web","target":"web:<user>:<session>","audienceScopeId":"..."}`. There is no destination type describing a human session on a surface other than web/Slack.
- `src/api/routes/keychain.ts` exposes credential/grant/ask CRUD plus `POST /v1/keychain/use`. There is no route that materializes a given user's granted credentials into a running sandbox.

The primitives already exist and are just not reachable: `Keychain` in `src/credentials/keychain.ts` declares `materialize`, `materializeOwn`, `materializeOwnFiles`, and `materializeStanding`, and every sandbox driver already calls `ephemeralCredLinkScript`. None of it can be driven from outside a turn.

The result: someone who reaches the same sandbox interactively — over SSH, or through a browser IDE attached to it — gets no credentials at all, even holding a valid `standing` grant for that scope.

## Why it matters

A sandbox belongs to a scope, not to a person, so its home directory is shared by everyone working in that scope and by the agent. Without per-session materialization the only options are:

1. Place a shared credential in the scope's home volume. That collapses every contributor into one identity on the git host, is readable by anyone who can reach the sandbox (the agent included), and has no expiry, no per-user attribution in `credential_usage`, and no revocation short of rotating the underlying token.
2. Ship no credentials, so nothing a human does interactively in the sandbox can authenticate.

[...]

PR #631, opened 2026-08-20
feat(keychain): materialize credentials for human sandbox sessions

yzxcj797:
Implements the suggested direction of #550 — the session-materialization half. The in-sandbox link plumbing (`ephemeralCredLinkScript`, `EPHEMERAL_CRED_DIR`) is reused exactly as the issue prescribes; what was missing was a non-agent driver, which this adds.

## What was unreachable

Every materialization path sat behind an agent capability token (`handleKeychain`'s gate) and ran inside a turn. The `Keychain` interface already declared `materialize`/`materializeOwn`/`materializeStanding`; every sandbox driver already ran `ephemeralCredLinkScript`; `renderUseScript` already produced the shell-sourceable form. A human reaching the same sandbox interactively (SSH, browser IDE) had exactly the two bad options the issue names: a shared credential in the scope's home volume, or nothing.

## The driver

**`Keychain.materializeStandingForUser(ownerId, scopeId)`** — `materializeStanding`'s selection filtered to one person's credentials. The grants table is keyed by audience scope, so the filter is what makes a shared sandbox's sessions per-user: Alice's session materializes Alice's grants, never Bob's, even though both granted toward the same scope. Once-mode and expired grants are excluded identically to the agent path.

**`POST /v1/keychain/sessions { userId, scopeId, ttlMinutes? }`** — source-auth (`auth: "source"`): the caller is the trusted integration holding `CORE_SIGNING_SECRET` (the SSH/IDE attach helper), naming the human on the session — the same trust boundary the issue's "authenticated internal route" describes. It materializes the user's set, records a `credential_usage` row per credential exactly as a turn does, audits `keychain.session.open`, and returns **the same `renderUseScript` output `/keychain/use` emits** — the helper sources it in the user's shell, so the materialization lifetime and isolation semantics are identical to an agent turn's. TTL bounded: default 8h, max 12h.

**`DELETE /v1/keychain/sessions/:id`** — idempotent release: `released` usage rows + audit. **`GET /v1/keychain/sessions`** lists live sessions for operators. An expired session sweeps to released on next touch — no re-materialization past its TTL.

**`Destination.humanSession { userId, sessionId }`** — the non-agent counterpart of the web destination, so asks and audit can name where a materialization went (the issue's item 1).

## What deliberately didn't change

The agent surface: existing routes, their capability gate, and turn-time materialization are untouched — the new routes are additive and source-auth-gated, so an agent token alone cannot reach them (and a source-auth caller was already trusted with the whole keychain surface these routes read).

## Tests

- `materializeStandingForUser returns only that person's standing grants` — Alice + Bob both grant toward one scope; each session gets exactly one credential (their own), a third user gets nothing. **Fails on `main`** (no such method).
- `materializeStandingForUser excludes once-mode and expired grants` — pins the mode boundary.

Full `keychain.test.ts`: 43 passed / 1 pre-existing environment failure (`sprites exec ... bad envelope` — identical on clean `main`); `keychain-ask` suite green; `tsc --noEmit` clean.

## Not included (per the issue's own scoping)

File-credential materialization for sessions (the agent path materializes files only via explicit `/keychain/use` re-fetch, and the same one-shot semantics deserve their own design), and the dashboard surface — the API + script contract here is the integration seam both would build on.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/631"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789831445&installation_model_id=19911&pr_number=631&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F631&signature=972eb622a1c4c773854af65a56219114f8a48e0f9abfa09efbf65e869d83464d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/api/routes/keychain.ts
+++ b/src/api/routes/keychain.ts
@@ -429,6 +429,135 @@ async function handleKeychain(ctx: ApiCtx): Promise<void> {
   return sendJson(res, 404, { error: "not_found" });
 }
 
+// ── Human-session materialization (#550) ────────────────────────────────
+//
+// A scope's sandbox is reachable interactively (SSH attach, browser IDE) by
+// the same people the agent serves, but credential materialization was only
+// drivable from an agent turn: the people got either a shared credential in
+// the scope's home volume (no expiry, no attribution, no revocation) or
+// nothing. These routes drive the SAME primitives (standing grants +
+// renderUseScript) for a human session, scoped to that one person's grants
+// and bounded by a session TTL.
+
+interface KeychainSession {
+  id: string;
+  userId: string;
+  scopeId: string;
+  createdAt: number;
+  expiresAt: number;
+  releasedAt?: number;
+  credentialIds: string[];
+}
+
+const keychainSessions = new Map<string, KeychainSession>();
+const KEYCHAIN_SESSION_DEFAULT_TTL_MS = 8 * 3_600_000;
+const KEYCHAIN_SESSION_MAX_TTL_MS = 12 * 3_600_000;
+
+function sweepKeychainSessions(): void {
+  const now = Date.now();
+  for (const s of keychainSessions.values()) {
+    if (s.releasedAt === undefined && s.expiresAt <= now) s.releasedAt = now;
+    // Retain released sessions briefly for idempotent DELETE / inspection.
+    if (s.releasedAt !== undefined && s.releasedAt + 3_600_000 < now) keychainSessions.delete(s.id);
+  }
+}
+
+async function handleKeychainSession(ctx: ApiCtx): Promise<void> {
+  const { res, deps, pathname, method, body, params } = ctx;
+  if (!deps.keychain) return sendJson(res, 404, { error: "not_found" });
+  const kc = deps.keychain;
+  sweepKeychainSessions();
+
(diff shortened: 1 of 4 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/765 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

Issue #630, opened 2026-08-20
Bot-id-form channel mentions (`<@B…>`) aren't recognized: top-level ones are silently dropped, in-thread ones misfire as ambient


ianTPE:
A member hand-typed `@qm` in a channel and got no response. Their client encoded the mention as the **bot id** `<@B0BND0YAX4J>` rather than the **bot user id** `<@U0BNGML9RHS>`. qm recognizes only the user-id form, and for the bot-id form Slack fires no `app_mention` event — the only path that dispatches a top-level channel mention — so the message is mirrored (`handled=f`) and dropped. Verified with a controlled debug-logging test; line refs against local HEAD (`f9aec5e`).

## Same bot, two ids

`bots.info` for `B0BND0YAX4J` returns `user_id: U0BNGML9RHS`, `app_id: A0BP7BCE0DN`, and that app has exactly one bot user in the workspace, so `<@B0BND0YAX4J>` is a legitimate mention of qm. Its `rich_text` block carries `{"type":"user","user_id":"B0BND0YAX4J"}` — Slack stored the mention pointing at the bot id, and the token survives entity-decoding and known-mention resolution in the mirror unchanged (`src/slack/mirror.ts:108`), so this isn't a mirror artifact.

## Controlled verification (`SLACK_LOG_LEVEL=debug`)

Two channel messages, same minute, debug-logging every inbound socket event:

| message | mention form | inbound event(s) | `app_mention`? | run? | result |
|---|---|---|---|---|---|
| `<@U0BNGML9RHS> …USERID` | bot **user** id | `message` **and** `app_mention` | yes | yes (`runs` row, done) | reply sent |
| `<@B0BND0YAX4J> …BOTID` | bot id | `message` only | **no** | none | `handled=f`, silent |

So in this workspace Slack fires `app_mention` only for the bot-user-id form (n=1 each, client-typed, private channel — observed behavior, not proven universal across channel types / API-posted messages).

## What actually happens to a bot-id mention

Dispatch of a top-level channel mention comes solely from the `app_mention` handler (`src/slack/events.ts:44`); the message handler mirrors the row and returns at `events.ts:140` for any top-level post. So:

- **B-form top-level mention → always dropped** (no `app_mention`, and the message path returns before dispatching).
[...]

PR #632, opened 2026-08-20
fix(slack): recognize bot-id-form mentions (<@B…>) everywhere

yzxcj797:
Fixes the recognition half of #630 — both silent drop and thread misdispatch. The message-handler dispatch of the B-form top-level mention (the exact class `app_mention` won't fire for) is the issue's step 2 and is left as a follow-up; this fix is its prerequisite and already stops the ambient misdispatch.

## What was missed

A hand-typed `@qm` can arrive encoded against the **bot id** (`<@BOTID>`) rather than the bot **user id** (`<@UBOT>`) — your controlled verification shows exactly this client behavior. Both forms are legitimate mentions of the same single-bot app (`bots.info` maps one to the other), but:

- Slack fires **no `app_mention`** for the B-form (your debug-log table), and
- `mentionsBot` recognized only the U-form,

so a top-level B-form mention fell through every gate (mirrored `handled=f`, no dispatch), and a B-form **thread reply** skipped the addressed-turn early return at `events.ts` — dispatching as an *unprompted ambient* turn with the raw `<@B…>` token still in the text.

## The fix — recognition at every site the U-form was read

- **`mentionsBot(text, botUserId, ownBotId)`** — both forms; the two-arg call shape is unchanged for every existing caller (your `threadHasBotStake` free-ride note holds: it forwards both ids now).
- **`stripMention(text, botUserId, ownBotId)`** — strips both forms *including the legacy `<@ID|label>` shape* (your regex note), so the core never receives the raw token on an addressed turn. Wired through `turn-handler`'s call site.
- **`events.ts`'s thread-reply gate and `mirror.ts`'s `mentionsSelf`** — both pass the bot id through; it was already in scope at both sites (`ids.ownBotId`).

Per your resolution warning, the bot id is **not** fed to `users.info` — no B→U translation exists there and the lookup would fail; display naming continues through the existing identity paths.

## Tests

- `mentionsBot recognizes the bot-id form (<@B…>) alongside the user-id form` — B-form, U-form, both, another bot, no ids configured, plus the two-arg back-compat shape. **Fails on `main`** (B-form reads false).
- `stripMention strips the bot-id form and the legacy label shape` — `<@BOTID>`, `<@BOTID|qm>`, `<@UBOT>`, `<@UBOT|qm>`, and an unrelated id left untouched. **Fails on `main`** (raw token survives).

Full `slack-message-gating` + `slack-mrkdwn`: 50/50 with the fix, 48/2 on `main` (exactly the two new tests); `slack-index.integration` + `slack-identity`: 67/67 unchanged; `tsc --noEmit` clean.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/632"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789834313&installation_model_id=19911&pr_number=632&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F632&signature=441fed8d96f0ac4e4dff59e688117719be9238736e35e12670a9caf6debc7413"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/slack/events.ts
+++ b/src/slack/events.ts
@@ -161,7 +161,7 @@ export function registerSlackEvents(
     if (m.channel_type === "channel" || m.channel_type === "group" || m.channel_type === "mpim") {
       if (m.channel_type === "mpim" && m.channel) syncForUnseenGroup(client, String(m.channel));
       const threadReply = isThreadReply(m);
-      const isMention = mentionsBot(m.text ?? "", ids.botUserId);
+      const isMention = mentionsBot(m.text ?? "", ids.botUserId, ids.ownBotId);
       const willDispatch = threadReply && !isMention && (await botHasStakeInThread(client, m.channel, m.thread_ts));
       await mirrorMessageEvent(m, client, willDispatch ? { handled: true } : {});
       if (!threadReply) return;
--- a/src/slack/message-gating.ts
+++ b/src/slack/message-gating.ts
@@ -1,7 +1,15 @@
 import { LRUCache } from "lru-cache";
 
-export function mentionsBot(text: string, botUserId: string): boolean {
-  return botUserId ? text.includes(`<@${botUserId}>`) : false;
+export function mentionsBot(text: string, botUserId: string, ownBotId = ""): boolean {
+  // Both id forms mention the bot: the bot USER id (`<@U…>`, the form
+  // Slack's autocomplete inserts) and the BOT id (`<@B…>`, which
+  // hand-typed mentions and some clients encode). A B-form mention is a
+  // legitimate mention of this bot — `bots.info` maps it to the same
+  // single-bot app — but Slack fires no `app_mention` for it, so without
+  // recognizing it here the message is dropped at top level and
+  // misdispatched as ambient in threads (#630).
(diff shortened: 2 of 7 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/764 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #634, opened 2026-08-20
fix(slack): dispatch bot-id-only mentions from the message handler

### LINKED ISSUE #630: Bot-id-form channel mentions (`<@B…>`) aren't recognized: top-level ones are silently dropped, in-thread ones misfire as ambient

yzxcj797:
Fixes the dispatch half of #630 (step 2 of the issue's fix sketch). **Stacked on #632** (the recognition commit — `mentionsBot`/`stripMention` both-id forms come from there); merge that first. Together they close #630 end to end.

## What still dropped after recognition

Even with `mentionsBot` recognizing the bot-id form, the message handler's channel path assumed `app_mention` would carry every mention: it returned before dispatch for top-level messages (`if (!threadReply) return`) and routed thread replies into the ambient stake-followed path (`unprompted: true`) when the mention wasn't the user-id form. Slack fires `app_mention` **only** for the bot-user-id form — the bot-id form had no dispatcher at all.

## The dispatch

The message handler now dispatches the **addressed** turn itself, scoped to exactly the class `app_mention` won't fire for:

```ts
const botIdMentionOnly =
  Boolean(ids.ownBotId) &&
  text.includes(`<@${ids.ownBotId}>`) &&
  !(ids.botUserId && text.includes(`<@${ids.botUserId}>`));
```

- **top-level bot-id-only mention** → dispatched as an addressed turn (previously: mirrored `handled=f`, silently dropped);
- **thread-reply bot-id-only mention** → dispatched addressed — `unprompted` is deliberately omitted: the sender typed `@qm`; the ambient flag belongs to the stake-followed path alone (previously: misdispatched as ambient carrying the raw `<@B…>` token);
- **both forms present** → the user-id form routes through `app_mention` as before; the scoping makes the two paths disjoint, so no double dispatch (pinned by a test that emits both events for the same ts and asserts exactly one turn);
- **neither form** → unchanged ambient/stake-followed behavior.

The mirrored row reads `handled` for every dispatched mention — the incident's `handled=f` signature for a message the bot actually answered is gone.

## Tests

- `a bot-id-form top-level mention dispatches an addressed turn` — turn dispatched, text stripped, **no** `unprompted`, mirror `handled`. **Fails on the base** (0 turns).
- `a bot-id-form thread reply dispatches ADDRESSED, not ambient` — `unprompted` undefined. **Fails on the base** (ambient dispatch).
- `a user-id mention still routes through app_mention, not the message handler` — both forms in one message, both events emitted, exactly one turn (disjointness).

Full identity suites (`slack-message-gating` + `slack-mrkdwn` + `slack-identity` + `slack-index.integration`): 120/120 on this stack; the two new dispatch tests fail on the recognition-only base (43/2). `tsc --noEmit` clean.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/634"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789841095&installation_model_id=19911&pr_number=634&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F634&signature=d33f56ebd04a9ea60a67a4a3894fbb9915d5fcebb0046586cf28e078f721c6e2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/slack/events.ts
+++ b/src/slack/events.ts
@@ -162,10 +162,23 @@ export function registerSlackEvents(
       if (m.channel_type === "mpim" && m.channel) syncForUnseenGroup(client, String(m.channel));
       const threadReply = isThreadReply(m);
       const isMention = mentionsBot(m.text ?? "", ids.botUserId, ids.ownBotId);
-      const willDispatch = threadReply && !isMention && (await botHasStakeInThread(client, m.channel, m.thread_ts));
+      // A bot-id-form mention (<@B…>) never triggers Slack's app_mention
+      // event — only the bot-user-id form does. Scoped to exactly that
+      // class (bot-id form present, user-id form absent), the message
+      // handler dispatches the ADDRESSED turn itself instead of dropping
+      // it at top level or mis-dispatching it as ambient in a thread
+      // (#630).
+      const text = m.text ?? "";
+      const botIdMentionOnly =
+        Boolean(ids.ownBotId) &&
+        text.includes(`<@${ids.ownBotId}>`) &&
+        !(ids.botUserId && text.includes(`<@${ids.botUserId}>`));
+      const willDispatch =
+        botIdMentionOnly ||
+        (threadReply && !isMention && (await botHasStakeInThread(client, m.channel, m.thread_ts)));
       await mirrorMessageEvent(m, client, willDispatch ? { handled: true } : {});
-      if (!threadReply) return;
-      if (isMention) return;
+      if (!threadReply && !botIdMentionOnly) return;
+      if (isMention && !botIdMentionOnly) return;
       if (!willDispatch) {
         console.error(
           `[slack-plugin] thread-follow skipped: no bot stake detected in thread ch=${m.channel} thread_ts=${m.thread_ts} ts=${m.ts}`,
@@ -191,7 +204,10 @@ export function registerSlackEvents(
           files: (m.files as SlackFile[]) ?? [],
           threadTs: m.thread_ts,
           ts: m.ts,
-          unprompted: true,
+          // A bot-id-only mention is ADDRESSED even in a thread — the
+          // sender typed @qm; the ambient flag belongs to the
+          // stake-followed path alone (#630).
+          ...(botIdMentionOnly ? {} : { unprompted: true }),
           ...(m.bot_id || m.subtype === "bot_message" ? { botAuthored: true } : {}),
           ackGate,
         },
--- a/src/slack/events.ts
+++ b/src/slack/events.ts
@@ -161,7 +161,7 @@ export function registerSlackEvents(
     if (m.channel_type === "channel" || m.channel_type === "group" || m.channel_type === "mpim") {
       if (m.channel_type === "mpim" && m.channel) syncForUnseenGroup(client, String(m.channel));
       const threadReply = isThreadReply(m);
-      const isMention = mentionsBot(m.text ?? "", ids.botUserId);
+      const isMention = mentionsBot(m.text ?? "", ids.botUserId, ids.ownBotId);
       const willDispatch = threadReply && !isMention && (await botHasStakeInThread(client, m.channel, m.thread_ts));
       await mirrorMessageEvent(m, client, willDispatch ? { handled: true } : {});
       if (!threadReply) return;
--- a/src/slack/message-gating.ts
+++ b/src/slack/message-gating.ts
@@ -1,7 +1,15 @@
 import { LRUCache } from "lru-cache";
 
-export function mentionsBot(text: string, botUserId: string): boolean {
-  return botUserId ? text.includes(`<@${botUserId}>`) : false;
+export function mentionsBot(text: string, botUserId: string, ownBotId = ""): boolean {
+  // Both id forms mention the bot: the bot USER id (`<@U…>`, the form
+  // Slack's autocomplete inserts) and the BOT id (`<@B…>`, which
+  // hand-typed mentions and some clients encode). A B-form mention is a
+  // legitimate mention of this bot — `bots.info` maps it to the same
+  // single-bot app — but Slack fires no `app_mention` for it, so without
+  // recognizing it here the message is dropped at top level and
+  // misdispatched as ambient in threads (#630).
+  if (botUserId && text.includes(`<@${botUserId}>`)) return true;
+  return Boolean(ownBotId) && text.includes(`<@${ownBotId}>`);
 }
 
 export function threadHasBotStake(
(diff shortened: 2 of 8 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/763 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #637, opened 2026-08-20
fix(slack): refresh identity caches on user_change/team_join events

### LINKED ISSUE #626: Email identity mode refuses an unresolved own-team member as an external guest, and the refusal can be fully silent

yzxcj797:
Fixes defect 3 of #626 — the last of the three. **Stacked on #628 (third state) and #629 (ok-propagation)**; the three together close #626 completely.

## The gap

Classification updates only on the 5-minute snapshot/lookup TTLs. Neither `user_change` nor `team_join` appears in the manifest's `bot_events` or anywhere in `src/slack` — so a just-set email kept an own-team member refused for the directory-propagation window (the incident's ~39 minutes was Slack-side propagation PLUS this TTL floor), and a profile change showed stale display names until a TTL lapsed.

## The fix

1. **Manifest**: `user_change` and `team_join` added to `bot_events`. *(Note for operators: the app needs a reinstall to receive the new subscriptions — the same README note the issue cites for scope changes.)*
2. **Handlers** (events.ts): both route the event payload into the new `Directory.updateUserFromEvent(user)` — the payload carries the **full fresh profile**, so the handler re-classifies from it directly (no `users.info` round-trip) and overwrites **both** caches the classifier reads:
   - the **LRU lookup entry** — the transient `classifyUserCached` path, including the `lastKnownClassification` fallback #629 added;
   - the **snapshot row** — the roster/membership resolution path.

   Both must move together or a change shows up in one surface and not the other until the TTLs lapse — exactly the split-state drift the issue's H1/H2/H3 diagnosis was blinded by.

Eviction (deleting the row) was deliberately not used: the event already holds the authoritative profile, so re-classifying from it is both immediate and cheaper than delete-then-refetch.

## Test

`updateUserFromEvent propagates a fresh profile into both caches (#626)` — email-mode fixture starting in the incident's exact shape (email-less member → `principalUnresolved: true`), then a `user_change` carrying the just-set email: the classification resolves **immediately** (`externalId` keyed on the fresh email), and the `lastKnownClassification` fallback serves the updated state, not the stale unresolved one. **Fails on the base** (no such method; the TTL-only path keeps the member unresolved).

Full identity suites: 30/30 on this stack; the suites' prior 120-test combined run stays 120/120 (`tsc --noEmit` clean).

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/637"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789842911&installation_model_id=19911&pr_number=637&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F637&signature=9d38792163f9ae73f71c9c6fbef9b2da4e4cc4879d76487a34499b33b8743ed4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/slack/directory.ts
+++ b/src/slack/directory.ts
@@ -66,6 +66,11 @@ export interface Directory {
   forceDirectorySync(client: any, invalidateChannelId?: string, invalidatePrincipalId?: string): Promise<void>;
   classifyUserCached(client: any, userId: string): Promise<CachedUser & { ok: boolean }>;
   lastKnownClassification(userId: string): ActorAssertion | undefined;
+  /** Apply a `user_change`/`team_join` event payload to the caches (#626):
+   *  re-classify from the fresh profile and overwrite BOTH the LRU entry
+   *  and the snapshot's row, so a profile/email change propagates
+   *  immediately instead of waiting out the TTL. */
+  updateUserFromEvent(user: SlackUser): void;
   classifyActor(
     client: any,
     userId: string,
@@ -599,6 +604,22 @@ export function createDirectory(deps: {
     }
   }
 
+  function updateUserFromEvent(user: SlackUser): void {
+    // The event carries the FULL fresh profile — classify from it directly
+    // (no users.info round-trip) and overwrite every copy the caches hold.
+    // The LRU entry is the transient-lookup cache; the snapshot row feeds
+    // roster/membership resolution. Both must move together or a name/email
+    // change shows up in one surface and not the other until the TTLs lapse
+    // (#626 defect 3).
+    const actor = classifyUser(user, ids.ownTeamId, ids.identityMode);
+    const timezone = slackUserTimezone(user);
+    const classified = { actor, ...(timezone ? { timezone } : {}) };
+    if (user.id) {
+      userCache.set(user.id, classified);
+      if (userSnapshot) userSnapshot.byId.set(user.id, classified);
+    }
+  }
+
   function lastKnownClassification(userId: string): ActorAssertion | undefined {
     // The LRU entry is the last SUCCESSFUL classification (failures are
     // never cached), i.e. exactly the fallback a transient users.info
@@ -679,6 +700,7 @@ export function createDirectory(deps: {
     classifyUserCached,
     classifyActor,
     lastKnownClassification,
+    updateUserFromEvent,
     getChannelInfo,
     channelMembership,
     allInternalRosters,
--- a/src/slack/directory.ts
+++ b/src/slack/directory.ts
@@ -65,7 +65,11 @@ export interface Directory {
   getUserSnapshot(client: any): Promise<{ byId: Map<string, CachedUser>; fetchedAt: number } | undefined>;
   forceDirectorySync(client: any, invalidateChannelId?: string, invalidatePrincipalId?: string): Promise<void>;
   classifyUserCached(client: any, userId: string): Promise<CachedUser & { ok: boolean }>;
-  classifyActor(client: any, userId: string): Promise<ActorAssertion>;
+  lastKnownClassification(userId: string): ActorAssertion | undefined;
+  classifyActor(
+    client: any,
+    userId: string,
+  ): Promise<ActorAssertion & { ok: boolean }>;
   getChannelInfo(client: any, channel: string): Promise<ChannelMeta | undefined>;
   channelMembership(
     client: any,
@@ -588,12 +592,28 @@ export function createDirectory(deps: {
       if (ids.ownTeamId) userCache.set(userId, classified);
       return { ...classified, ok: true };
     } catch {
-      return { actor: { externalId: userId, isExternalGuest: true }, ok: false };
+      // Lookup failed — say so instead of asserting the user is an external
+      // guest (#626): the id-based actor carries lookupFailed so consumers
+      // can distinguish "couldn't check" from every checked state.
+      return { actor: { externalId: userId, lookupFailed: true }, ok: false };
     }
   }
 
-  async function classifyActor(client: any, userId: string): Promise<ActorAssertion> {
-    return (await classifyUserCached(client, userId)).actor;
+  function lastKnownClassification(userId: string): ActorAssertion | undefined {
+    // The LRU entry is the last SUCCESSFUL classification (failures are
+    // never cached), i.e. exactly the fallback a transient users.info
+    // failure should fall back to (#626). Time-bounded by the cache TTL.
+    return userCache.get(userId)?.actor;
+  }
+
+  async function classifyActor(client: any, userId: string): Promise<ActorAssertion & { ok: boolean }> {
+    // ok preserved, not discarded: a failed users.info currently reads as a
+    // confident "external guest" to every caller — transiently refusing a
+    // member whose record is merely unreachable. Callers decide what a failed
+    // lookup means (skip, retry, fall back to last-known); the classification
+    // itself must not launder the failure into an identity claim (#626).
+    const { actor, ok } = await classifyUserCached(client, userId);
+    return { ...actor, ok };
   }
 
   async function getChannelInfo(client: any, channel: string): Promise<ChannelMeta | undefined> {
@@ -658,6 +678,7 @@ export function createDirectory(deps: {
     forceDirectorySync,
     classifyUserCached,
     classifyActor,
+    lastKnownClassification,
     getChannelInfo,
     channelMembership,
     allInternalRosters,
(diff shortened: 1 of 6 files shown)


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/761 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #654, opened 2026-08-21
feat(slack): render agent bodies as native markdown blocks

### LINKED ISSUE #651: Slack replies are mrkdwn text, so lists have no structure and CJK bold prints asterisks
Every agent reply leaves the Slack surface through `toSlackMrkdwn` and is posted as `text` (`turn-handler.ts:542`, `delivery.ts:43`). mrkdwn has no list, table or heading primitive, so structure is encoded as characters: a bullet becomes a literal `•`, a GFM table an aligned code block, a heading bold. It reads as a list until a line wraps, and then the second line returns to column 0.

The same pass turns `**x**` into `*x*` (`mrkdwn.ts:87`), and single-asterisk emphasis rendered only where both delimiters sat against a half-width boundary. In a CJK sentence they rarely do, so a Chinese-language reply cannot reliably use bold.

Slack's `markdown` block takes standard Markdown and translates it into native blocks. I sent one sample twice to the same conversation: **A** as `text` only (today's path, the string produced by this repo's own `toSlackMrkdwn`), **B** as the same `text` plus `blocks: [{ "type": "markdown", "text": <raw Markdown> }]`. Then read both back with `conversations.history` and walked the element tree.

| construct | A: mrkdwn `text` | B: `markdown` block |
|---|---|---|
| bullet list | no `rich_text_list` element; all three levels are the same literal `•`, separated only by leading spaces; a wrapped line returns to column 0 | `rich_text_list style=bullet` at `indent=0`, `1`, `2`; distinct marker per level; wrapped lines hang-indent |
| ordered list | passes through as literal text (`test/slack-mrkdwn.test.ts:81` asserts this) | `rich_text_list style=ordered` |
| `**x**` flanked by CJK | literal `*x*`; of the sample's three bold positions only the one flanked by half-width spaces rendered | all three rendered as bold spans |
| GFM table | aligned code block; `padEnd` counts a CJK character as width 1, so any CJK column misaligns | native `table` block |
| fenced ` ```python ` | `rich_text_preformatted` with no `language`; `python` becomes the block's first line of content | `rich_text_preformatted` with `language: "python"` |
| `<@U…>` mention | resolved | resolved, native `user` element |

**A, what the plugin sends today:**

Image

**B, the same body in a `markdown` block:**

Image

## Why

Structure is encoded into characters in `toSlackMrkdwn`:

```
:67   fenced blocks pass through untouched, so the info string lands in a
      plain-text field as content
:87   **x** / __x__  →  *x*
:97   -, *, + at line start  →  "• "
:158  GFM table  →  aligned monospace code block
```

Ordered lists are not handled at all. `blocks` is only ever set for the task-list surface (`presenters.ts:269`, `deliveries.ts:127`) and the approval cards, and those are `section` blocks carrying mrkdwn, so they share the limitation.

Six sites render an agent-authored body this way, so a change has to cover all of them: `turn-handler.ts:325` (first-block ack) and `:542` (the reply), `approvals.ts:239` and `:648`, `deliveries.ts:105` and `:233` (the delivery poller, which is how cron and background results reach Slack).

## Measured about the markdown block

- One block came back as three (`rich_text`, `table`, `rich_text`), so the per-payload block limit applies to the translated result, not to what you send.
- A `section` block sent alongside it survived unchanged, so the task-list surface can keep its section block.
- With `blocks` present Slack rewrote `text`, turning newlines into spaces. Without `blocks` it came back byte-identical.

## Implementation notes

1. The block has to carry the raw Markdown, not `toSlackMrkdwn`'s output: `*x*` is italic in Markdown and `<url|label>` is not link syntax. The raw body is already in scope (`replyBody`, `turn-handler.ts:533`).
2. Keeping the mrkdwn string as `text` still earns its place as the notification and screen-reader fallback, and `mirrorSelfPost` passes it to the surface cache the web UI reads.
3. The budgets do not line up: `safeChunks` at 2,900 per section block (`mrkdwn.ts:176`), `SLACK_TEXT_LIMIT` 40,000 for `chat.update`
   (`messaging.ts:18`), and 12,000 cumulative per payload for markdown blocks. I did not test the last two.
4. Splitting long Markdown safely is a different problem from splitting mrkdwn safely: `safe-cut.ts` guards `<url|label>` and unbalanced `*` / `~` runs, while Markdown needs fences, table rows and list continuity guarded, and a cut inside an ordered list restarts the numbering. Emitting the block only when the body  is under budget, and falling back to today's path above it, keeps that off the critical path.
5. `updateSlackMessage` (`messaging.ts:28`) passes `blocks: blocks ?? []`, and an empty array clears blocks, so the edit paths have to resend the markdown block  or a message reverts to plain text on its first update.
6. `test/live-slack/scenarios.ts:90` asserts three lines beginning with `•`. A native list puts no `•` in the text layer, and the text field is rewritten anyway, so that assertion needs to look at block structure instead.

## Fix

Render each outbound body once into `{ text, blocks }`: `text` from the existing `toSlackMrkdwn`, `blocks` a single `markdown` block carrying the raw Markdown, omitted when the body is over budget. Then switch the six sites to it. `postReply`, `chat.update` and the task-list presenter already accept `blocks`.

yzxcj797:
## Summary

Implements the fix shape from #651 verbatim: every outbound agent body renders once into `{ text, blocks }` — `text` stays the existing `toSlackMrkdwn` fallback (notifications, `mirrorSelfPost` cache, older clients), and a single `markdown` block carries the **raw Markdown**, where lists, tables, fenced-code languages and CJK-flanked bold translate into native Slack elements instead of character-encoded approximations.

## Implementation

- `renderSlackBody(md)` in `mrkdwn.ts` returns both renderings; `slackMarkdownBlocks` wraps the raw Markdown. The block is emitted only when the body fits `SLACK_MARKDOWN_BLOCK_BUDGET = 11,500` — Slack's 12,000 budget applies to the *translated* result cumulatively, so over-budget bodies fall back to today's text-only path, keeping Markdown-safe chunking (fences, table rows, ordered-list continuity) off the critical path per the issue's note 4.
- All six agent-body sites render through it:
  - `turn-handler.ts` — the ack presenter and the turn reply (`postReply` already accepted `blocks`);
  - `approvals.ts` — both result paths; the non-remote card update **resends** the block (an update with no blocks clears them, per note 5), and `postApprovalFollowup` threads it; the agent-to-agent completion header is bolded `**` for the block and stays `*` in the text fallback, since `*x*` is italic in Markdown (note 1);
  - `deliveries.ts` — both pollers; the channel poller's markdown block replaces its section-block body while the task-list `section` block survives alongside (per the "a section block sent alongside it survived" measurement), and the principal poller posts it directly.
- The `text` fallback path is byte-identical to today, so `test/slack-mrkdwn.test.ts`'s existing expectations are unchanged.

## Testing

- 4 new tests: the helper's contract (raw Markdown in the block, mrkdwn in text; over-budget and blank bodies fall back to text-only; the budget stays under the 12k cap) and a wiring contract asserting every one of the six sites threads the blocks. Differential: with the source changes stashed the suite fails to import; neighbors (`slack-delivery`, `slack-approval-cards`, `slack-ack-emoji`) are 72/72.
- Note 6's live-scenario assertion (`test/live-slack` asserting `•` lines) still applies to the `text` layer, which is unchanged; the native-list layer only appears through the block, which the live suite would need block-structure assertions for — left to that suite's maintainers since it requires a live workspace.

Fixes #651

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/654"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789922963&installation_model_id=19911&pr_number=654&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F654&signature=395d9a8d85ffa217ecdfac0c1cc90b75f64ff977cfd856a0ff722feb47b98726"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


time-attack:

We don't want your idea to get lost, so we've captured it in https://github.com/yc-software/qm/issues/752 to triage on our side. Please feel free to open an `adrs/` note if you feel strongly about the approach. Appreciate you engaging with the repo!

PR #706, opened 2026-08-27
Cloud browsing via Browser Use: admin key setting, browser_use tool, live view in web UI

time-attack:
## What

Admins can enable cloud browsing for agents by pasting a single Browser Use Cloud API key in the admin UI. When set, agents get a `browser_use` tool that runs web tasks on Browser Use's hosted browser agent, and the person watching in the web UI sees a **collapsible live view of the actual browser** (interactive — they can take over, e.g. to complete a sign-in) embedded on the running tool row.

- **Admin setting** (`browser-use-key`, org-scoped): encrypted with the connector key, Postgres-backed, write-only, live-validated against the provider on save, explicit Clear button (no save-empty footgun). Reads back only a configured/not-configured flag derived from decryptability, so key-material rotation can't show "Configured" while agents silently have no tool.
- **`browser_use` tool** (all four harnesses, threaded through `HarnessTurnInput` like `credentialExecServices`): `POST /api/v4/runs`, polls to completion, 15-minute wall clock with remote cancel, abort-safe (mid-poll aborts still cancel the paid run; creation ignores the caller signal so an abort can't orphan an uncancellable run), tolerant of transient poll failures, defensive on unexpected statuses. Completed **and** failed outputs route through the external-content security screen. Offered only on non-read-only, all-internal turns with an internal actor; strict-posture approval gating applies.
- **Live view**: the tool re-emits its `tool_call` with the live-view URL (origin-validated to `https://*.browser-use.com` in core); the timeline merges same-callId re-emits into one row; the web UI renders an open collapsible sandboxed iframe (`sandbox`, `referrerpolicy=no-referrer`, https-only guard) while running, plain row once settled so stale URLs never reload on replay.

## Deliberate scope choices (called out for review)

- No per-scope ACL grants or admin-tunable spend ceiling yet — the org key is the opt-in and the 15-minute wall clock is the cap. Strict posture still gates each call.
- Unattended trigger/cron fires may use the tool, matching the existing browse skill's posture.
- The cloud browser egresses from Browser Use's network, outside org egress policy; the admin card copy discloses this.
- Dedicated key storage (config-store map) rather than reusing ServiceCredentialStore: the key is consumed core-side (never delivered to sandboxes), wants live validation + a first-class admin card, and scope-config PUTs are already audited.
- lit's non-keyed message list can recreate the iframe on reattach/history-expand; that reloads the *view* only — the remote browser session and any in-progress takeover state live server-side.

## Review

Reviewed pre-PR by an 8-angle fresh-context pass; all confirmed findings fixed, including: the tool never registering on the pi harness (tool list is built before `ref.current` exists — now threaded through harness options), unscreened failure text, deferred `tool_call` breaking replay/turn-resume accounting, iframe sandbox/scheme hardening, admin "Configured" vs undecryptable divergence, and orphaned paid runs on mid-flight aborts.

## Testing

- 239 core tests green locally (pi-tools, browser-use client, admin resource route incl. org-only + validation-rejection + clear, config-store encryption/restart/rotation, orchestrator; Postgres persistence parity runs in CI with DATABASE_URL).
- web-ui: timeline merge unit tests + live-view source assertions (31 green). admin: card/save/clear assertions (80 green). Typecheck + lint clean everywhere.
- **Not exercised end-to-end**: a real Browser Use run needs a live API key, which this environment doesn't have. The no-key state is a strict no-op (no tool, card shows "Not configured"). First live QA will happen on the deploying instance right after sync.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/706"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790465240&installation_model_id=19911&pr_number=706&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F706&signature=cda21a7ae1e6243529bfe3832032e5b0e8839fd43f2479ae611838b5f4556a80"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MagMueller:
sina, this is sick. happy to help test the live Browser Use path before merge.

PR #690, opened 2026-08-26
Subscription auth: Codex ChatGPT OAuth + harness-carried model auth (supersedes #126, #128)

ReganBell:
Adds subscription-backed Codex auth and recognizes harness-carried credentials without treating them as general provider API keys.

- verification: `npm run typecheck`
- verification: formatting and focused lint
- verification: 111 focused core tests
- verification: 4 CLI regression tests

PR #707, opened 2026-08-28
Individual model auth: per-user Claude/ChatGPT accounts drive each turn

time-attack:
Updates and extends the `subscription-auth` work in #690 (this machine cannot push to that PR's head branch, so this PR carries the full branch from a fork; #690 can be closed in its favor, or a maintainer can fast-forward its branch to this head).

## What this adds

An org-wide **"Individual authorization for AI usage"** toggle (admin → Governance). When on, every user connects their **own** AI account before chatting, and each of their turns runs on that account — never the org's shared credentials:

| User connects | Turn runs on | Credential path |
|---|---|---|
| Claude subscription (PKCE sign-in) | `claude` harness | `CLAUDE_CODE_OAUTH_TOKEN` injected per turn |
| ChatGPT subscription (device-code sign-in) | `codex` harness | per-turn `auth.json` written into the jail; refreshed tokens persisted back to the user's store |
| Pasted Anthropic/OpenAI API key | `pi` harness | key resolved per turn via `turn.providerKeys` |

## Pieces

- **Store**: `user-model-credential-store` — encrypted (existing connector-secret crypto), keyed `(user, provider)`, over the durable artifact map (`user_model_credentials`).
- **OAuth**: `subscription-oauth` — Claude PKCE (claude.ai authorize + console token exchange) and ChatGPT device-code (auth.openai.com), refresh for both, 15s timeouts on every call.
- **Routing**: `individual-auth-routing` (pure, unit-tested) picks provider/harness/model from the user's credential; orchestrator refreshes OAuth before the turn and threads `claudeOauthToken` / `codexAuth` (+ refresh write-back callback) on `HarnessTurnInput`.
- **Harnesses**: claude harness overrides the child env token per turn; codex harness writes a per-user sanitized `auth.json` under the existing turn auth lock and persists refreshed tokens to the per-user store instead of the shared file.
- **API**: `/v1/user-model-auth/*` (status, api-key, disconnect, claude start/complete, chatgpt start/poll), all user-scoped-enforced; web-ui server proxies under `/api/user-model-auth/*` and `/me` carries `individualModelAuth` + `modelAuthConnected`.
- **Web UI**: first-login connect gate + self-serve manage panel (sidebar footer ✨): per-provider rows with real brand marks, subscription sign-in or API key per provider, click-to-copy device code, disconnect to switch; composer's model/harness picker hidden when the account decides the runtime.
- **Pre-flight fix**: the deployment-level "provider isn't configured" refusal is skipped for individual-auth web turns, and client-sent harness/model are ignored — the connected account decides.

## Safety
- Org/shared credential paths are unchanged when the toggle is off (default off).
- Per-user turns resolve keys exclusively from the user's store — `buildModelRuntime` receives only `turn.providerKeys`, so an unconnected user cannot ride the org key.
- Secrets encrypted at rest; API keys validated against the provider before storing; audit events on connect/disconnect.

## Testing
- `test/user-model-auth-injection.test.ts`: store round-trip, routing matrix (apikey→pi, anthropic-oauth→claude, openai-oauth→codex, requested-provider precedence, no-cred→null), codex `auth.json` validity.
- Affected suites green: codex/claude harness, harness-adapter, model-credential-route (82/82), web-ui 565/565, admin suite, typecheck + eslint clean across core and plugins.
- QA'd live on a dev instance: admin toggle round-trip, gate, both OAuth starts against the real providers, api-key validation, per-turn routing verified in core logs (`[individual-auth] user=… harness=… auth=…`).

## Demo
Exercised end-to-end in a local dev instance (portal → web-ui → core). Screenshots of the connect gate/manage panel to follow in a comment — the panel is the `✨ AI account` button in the sidebar footer with individual auth on.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/707"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790469966&installation_model_id=19911&pr_number=707&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F707&signature=3e174150c3a7db5e967b8b96afd112532d8ed7fc2da0262d5a0fef8175c6d495"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


time-attack:
A high-effort fresh-context review (8 finder lenses + adversarial verification) ran against this diff and confirmed 10 findings; all are resolved in 80f74af:

**Auth enforcement (the serious ones)**
- Individual auth now **fails closed**: a human turn with no connected account is refused with a connect prompt — previously it silently ran on the org's shared credentials whenever the org default harness was claude/codex (and Slack has no gate). Automation/cron/ambient turns intentionally bypass individual auth and stay on org credentials, so enabling the toggle no longer breaks or personally-bills scheduled work.
- The per-user harness override is **pinned through the router**, so it is no longer rejected by the approved-harness list (which previously hard-failed every turn for connected users in default-configured orgs).
- Claude subscription turns **strip the org's ANTHROPIC_API_KEY/ANTHROPIC_AUTH_TOKEN** from the child env — before, Claude Code preferred the org key and silently billed the org while logging auth=claude-oauth.
- The codex jail auth.json is **restored to org state after every per-user turn** (it previously persisted, so later org turns in API-key orgs ran on the user's personal ChatGPT account), per-user and org turns now **serialize on the same lock**, and the token write-back is guarded by lock ownership + account-lineage verification so concurrent turns cannot cross-contaminate stored credentials.
- Disconnect is a **hard delete** and refreshed tokens re-check connection state before persisting, so a disconnect cannot be resurrected by an in-flight turn and revoked secrets don't linger.

**Gates & UX correctness**
- /me fails closed (503) when auth status is unavailable instead of letting unconnected users bypass the gate; its three core calls now run in parallel.
- The ChatGPT device poll stops at code expiry with a retry message instead of polling forever.

**Reuse/cleanup**
- Provider-key validation reuses the admin validator (keeping providerBaseUrl overrides — user keys now validate correctly in proxied deployments); PKCE and ChatGPT-JWT helpers reuse the existing connectors/codex helpers; status is a single `connections` shape read without decrypting secrets; errMessage in route catches; removed the duplicate flush-key spread in config-store and dead store methods.

Verified after the fixes: core affected suites 82/82, web-ui 565/565, typecheck+lint clean, and a live dev-instance turn on a real Claude subscription routed `harness=claude auth=claude-oauth` end to end.

Known deferrals (structure, not correctness): consolidating the three turn-auth fields into one discriminated union on HarnessTurnInput, extracting the codex lock-acquire loop shared by the org and per-user branches, and /:provider-style OAuth route naming.

time-attack:
Merged @reganbell's keychain-custody commit (c1dd286 from #690) into this branch and re-expressed the per-user paths on its custody model — this PR now carries both bodies of work as one:

- **Per-user Codex turns ride `CodexAuthStore`'s primitives**: the user's tokens are validated with `codexOAuthAuthFromValue` and materialized as refresh-token-free child auth via `prepareCodexHome`; the jail auth is restored to org state after the turn. The write-back/lock/lineage machinery this branch previously carried for per-user Codex is deleted — the orchestrator already refreshes and persists per-user tokens *before* the turn, which is exactly the custody model.
- **Identity isolation without the old file locks**: once individual auth has been used, Codex turns that touch OAuth or per-user auth serialize through one in-process slot, so concurrent turns from different identities can't run on each other's accounts. API-key orgs that never use individual auth keep full concurrency (matching c1dd286's behavior).
- **Per-user Claude turns layer the user's token over the keychain `authEnv`** material, stripping org API keys from the child env.

Verified after the merge: typecheck + lint clean, harness/auth suites 76/76 (one known-flaky codex cancellation test passes in isolation and on rerun), and a live dev-instance turn on a real Claude subscription routed `harness=claude auth=claude-oauth` end to end.

Open custody question for @reganbell: per-user subscription tokens currently live in the new `user_model_credentials` store (encrypted via the connector key). If you'd rather they be keychain credentials so there's a single custody home, that's a contained follow-up — the orchestrator's refresh-before-turn shape stays the same either way.

With c1dd286 now contained here, #690 can be closed in favor of this PR.

ReganBell:
Pushed `fb54c46` (point 1 from review): per-user AI-account custody is now unified onto the keychain instead of the parallel `user_model_credentials` store. Same `UserModelCredentialStore` interface, so routing/orchestrator/API/UI are untouched — but each (user, provider) login is now an ordinary keychain credential (service `model-anthropic`/`model-openai`, origin `individual-model-auth`): keychain encryption, ownership checks, admin visibility, and "remove my credentials" cover AI logins for free. Added `Keychain.readOwnSecret` (owner-only decrypt, no grant machinery). Token expiry stays inside the payload so an expired access token still surfaces its refresh token to the pre-turn refresh. Tests extended to prove unified custody; typecheck/lint/prettier clean; keychain + codex + model-credential suites green. Remaining review points (shared-jail serialization, refresh single-flight/CAS, refresh token crossing the harness boundary) still open.

PR #848, opened 2026-08-29
local sandbox: recreate a parked container whose docker network is gone

irisfeng:
## Problem

Each local-sandbox container runs on its own docker network and is parked (`docker stop`) between turns. Docker records that network in the container's endpoint config **by ID**. If the network is removed out-of-band while the container is parked — a manual `docker network prune`, a daemon restart that loses networks — the container can never start again:

```
Error response from daemon: failed to set up container networking: network <id> not found
```

`ensureContainer`/`ensureScratch` saw a stopped container with a matching image and went straight to `docker start`, which throws that error verbatim on every subsequent session. The only recovery was removing the container by hand. Recreating the network under the same name does not help, because the stored endpoint references the old network ID.

## Fix

Before reusing a **stopped** container, verify the networks it still references exist (`networksIntact`: inspect the endpoint map, check each `NetworkID`). When one is gone, fall through to the existing recreate path — `rm -f` + `docker run` on a freshly created network. That path remounts the per-scope named volume, so home-directory data survives. Running containers short-circuit past the check (their networks cannot be pruned out from under them).

The scratch path gets the same guard, plus the `rm -f` its fall-through previously never needed (reaching it used to imply the container did not exist).

Deliberately unchanged: `ensureRunning` (restart under a live handle) still surfaces the raw `docker start` error. That scenario implies the daemon died mid-session — execs fail regardless — and every new session enters through `provision`, where the guard now breaks the wedge permanently.

## Tests

Two new cases (scoped + scratch): park a container, delete its network, provision again → container recreated on a fresh network, same volume remounted, warm home (`coldStart: false`). The fake docker now models the failure mode itself — `start` fails with the "network not found" daemon error when the endpoint's network is missing — so both tests fail on the base commit with exactly the user-facing error this PR fixes, and pass with it. Full `test/local-sandbox.test.ts` suite passes (20/20), tsc and oxlint clean.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/848"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790608835&installation_model_id=19911&pr_number=848&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F848&signature=b47c2d638b169cff80532176c5e5e2556172fd3555a451565c7737e35105bb0a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/sandbox/local-sandbox.ts
+++ b/src/sandbox/local-sandbox.ts
@@ -146,6 +146,21 @@ export function createLocalSandbox(workspace: WorkspaceStore, opts: LocalSandbox
     return { running: running === "true", imageId };
   }
 
+  async function networksIntact(name: string): Promise<boolean> {
+    const r = await dexec(["inspect", "-f", "{{json .NetworkSettings.Networks}}", name]);
+    if (r.code !== 0) return false;
+    let endpoints: Record<string, { NetworkID?: string }>;
+    try {
+      endpoints = JSON.parse(r.stdout.trim() || "{}");
+    } catch {
+      return false;
+    }
+    for (const endpoint of Object.values(endpoints ?? {})) {
+      if (endpoint.NetworkID && (await dexec(["network", "inspect", endpoint.NetworkID])).code !== 0) return false;
+    }
+    return true;
+  }
+
   async function resolvePort(name: string): Promise<number> {
     const cached = portByName.get(name);
     if (cached) return cached;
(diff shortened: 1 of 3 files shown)


irisfeng:
Friendly ping for review — happy to adjust scope or split the change if that helps. The failure mode (a parked container whose docker network was reclaimed) hits any deployment that prunes networks, and the PR ships with tests covering the exact user-facing error. Anything needed from my side to move it forward?

PR #872, opened 2026-08-31
memorable: optional procedural memory, off by default

NIkhil-cmd-cmd:
Gives QM procedural memory. When a session ends, its tool-call trace is parsed,
deterministically, with no model in the pipeline, into a procedure recording
which files changed and what verified the work. A later session on a similar
task gets a short pointer appended to its prompt and goes straight to the fix
instead of re-deriving the diagnosis.

Rebased onto current `main`, **0 deletions**. Supersedes the five stacked PRs
(#1, #2, #3, #4, #5).

## Three switches, and none of them is on

`MEMORABLE=1` enables the integration; `QM_MEMORABLE=0` is the kill switch that
wins over everything. Both are parsed once in `loadConfig`, per the
config-boundary rule. The relay also honours QM's own `memoryPolicy`: with
capture off no session-end hook is registered, and with recall off no injection
is attempted, so procedural memory cannot outlive the switch operators already
have.

Then, separately, **connecting is not consent**. Nothing is captured for a scope
until someone sets it to `read-write`. Two absent answers are `unset`, and
`unset` is deny.

## Each person connects their own account

One key per deployment is the wrong shape for a multi-tenant harness: every
scope's procedures land in one organization, and whoever holds that key reads
all of them.

So a person connects their own, through a device authorization (RFC 8628). QM
asks the sign-in service for a code, hands them a URL, and stores whatever key
comes back, encrypted. QM never sees a password and cannot create an account for
someone who has not signed in themselves.

```
POST   /v1/memorable/connect     start it; returns a URL and a code to show the person
GET    /v1/memorable/connect     poll; `connected` once they approve
DELETE /v1/memorable/connect     forget the key and any sign-in still in flight
POST   /v1/memorable/consent     {mode}: read-write, read-only, or deny
GET    /v1/memorable/accounts    admin-only; which scopes connected, never the keys
```

**These act on the caller's own `personal:<actorId>` scope and nothing else.**
Naming any other scope is refused, with no admin override, for the same reason
`connectors.ts` refuses it: whoever opens the URL is whoever the key belongs to,
so binding it to a channel would file the first person who clicked under
everyone else's name, and binding it to another person's scope would let an
admin route their transcripts to an organization they do not control.

The consequence, which is deliberate: a session in a shared channel runs under
`channel:<id>`, which nobody can connect, so it uses the deployment's own key.
Per-person accounts cover personal-scope sessions. A key is never borrowed by a
scope that did not connect it.

## Credentials

Stored keys are encrypted at rest with
`deriveConnectorKey(CONNECTOR_SECRET_KEY, "memorable-accounts")`, the same
AES-256-GCM path `model_credentials` uses. Without that key material the account
store is not built at all, rather than writing a bearer token to Postgres in the
clear. A row that will not decrypt reads as no key rather than throwing.

Spawned children get an allow-listed environment built once in `loadConfig`,
mirroring `codexProcessEnv` and `claudeProcessEnv`. `DATABASE_URL` is
deliberately **not** in it: the child gets the connection string only as
`MEMORABLE_DB_URL`, the one name meant for it. `MEMORABLE_BACKEND` defaults to
`qm` so the CLI cannot be silently misconfigured into writing a per-machine file
instead of your Postgres.

## Egress, stated plainly

An earlier revision of this branch claimed QM makes no network call of its own.
With accounts that is no longer true, and the docs say so rather than burying
it. QM calls two endpoints, only when someone starts a connect:

| Endpoint | Carries |
|---|---|
| `POST <api>/v1/device/code` | a label like `qm personal a1b2c3d4`: the scope kind and a truncated hash, never the scope id |
| `POST <api>/v1/device/token` | the opaque device code |

Neither carries a prompt, a tool call, a file, or a transcript. `<api>` is
`MEMORABLE_API_URL`, so you can point it at your own deployment. The `fetch` is
injected, so no test here reaches the network. Recall and capture stay
spawn-only.

## Where the data goes

Five tables on QM's DurableMap row shape, in the database `DATABASE_URL` already
points at. No new database and **no migration**: three are the CLI's, created on
first write, and two are QM's own, created lazily by `artifactMap` exactly as
`consent_links` and `secret_drops` are.

```
memorable_procedures  memorable_mode  memorable_stats      (the CLI's)
memorable_accounts    memorable_device_codes               (QM's)
```

Pulling the integration out leaves all five behind; drop them if you want the
data gone. `docs/procedural-memory.md` says this rather than leaving it to be
found in `psql`.

## What review changed

Two independent reviewers went at this with no part in writing it. Everything
below is a bug they found and this branch fixes, with a test each.

- **A disconnect during an in-flight approval left a live key behind.** `poll()` read the pending record, spent up to ten seconds on the network, then wrote the account without rechecking. A `DELETE` landing in that window reported `disconnected: false` and the key was persisted anyway, so that scope's transcripts kept flowing to an org nobody thought was connected. `poll()` now claims the pending row with `DurableMap.take()` before writing, and a lost claim reports the truth instead.
- **Any admin could bind any scope.** `authorizeAdmin` ignores its scope argument, so it is a global admin check, and these routes sit outside `/v1/admin/` where the live-actor and DM guards run. An injected prompt in an admin's cron turn could have rebound a victim's scope. Now self-only, matching what `connectors.ts` already decided for the identical operation.
- **The relay ignored `memoryPolicy`.** A deployment running capture off still relayed full transcripts.
- **`DATABASE_URL` went to a third-party binary** on every turn, alongside a redundant `MEMORABLE_DB_URL`. Removed.
- **`apiKey` without `env` silently dropped the key** and inherited the whole process environment. The type no longer allows it.
- **The API key was truncated at 200 characters** and stored corrupt. Bounded and rejected instead.
- **Every non-200 from the token endpoint read as "waiting for approval",** so a 5xx outage or a 429 looked identical to a human who had not clicked yet, and after ten minutes the operator was told `expired`. Now 5xx and 429 report `unavailable` and keep the code alive; a 4xx retires it.
- **A second `start()` orphaned the code the person was already looking at.** It now hands back the live one.
- The org fallback tier was unreachable once connects went self-only, so it is gone rather than left as dead code.

Reviewers also confirmed clean: the encryption purpose and ordering, no key in
any response, audit record, log or catalog, the device-code label, spawn argv
safety, and that RAM-backed maps without `DATABASE_URL` are the repo-wide
pattern rather than something this branch introduced.

## Known gaps

- **Capture is not deeply redacted.** `captureSession` forwards tool inputs and the prompt, capped, through the CLI's own minimization. QM does not run `createSecretValueMasker` over them. A credential pasted into a tool argument would travel. This should be settled before anyone turns capture on for a scope that handles secrets, and I would rather agree the approach with you than guess at it.
- **No QM-side rate limit** on connect or poll. The sign-in service limits per IP, which for a deployment is one shared egress IP.
- **No sweeper for `memorable_device_codes`.** A scope that starts a sign-in and never polls leaves its row until it does.
- **The relay has no watermark**, so a long thread re-offers its earlier workflows at each run end. The CLI skips what it has already stored, so this is wasted spawns rather than duplicate rows.
- Three of the four connect routes are capability-only in practice, so the portal cannot drive them even though they are declared `auth: "either"`.

## Verification

Typecheck, eslint, prettier and knip clean. `npm test` green.

79 tests across the six memorable files. The ones that would have caught the
list above: a disconnect racing an in-flight approval leaves no key; two
concurrent polls both report the truth; a 502 keeps the code and a 400 retires
it; an over-long or missing key stores nothing; a key is never borrowed by
another scope; the stored row never carries the key in the clear; a row under
different key material reads as no key rather than crashing; and every connect
route refuses a scope that is not the caller's, in the body and in the query
string.

The device flow was also round-tripped against the live service using
`accounts.ts` itself rather than a stub: `start` returned a real approval URL,
`poll` returned `pending`, `keyFor` returned null until approval.

The red checks on this PR are the fork, not the branch. This mirror has no
Actions runners, so every job fails in three to five seconds with no steps and
no runner assigned. Unmodified upstream `main`, pushed to this same mirror,
fails identically.

## Numbers

Measured on QM's own bench, codex harness, local docker sandbox, live
extraction path, pass decided by an independent in-sandbox `sh test.sh` run by
the harness rather than the agent's claim.

**Tool calls drop 40%**, pooled over three independent replications at n=25 per
arm (150 runs), Mann-Whitney z=4.06, p=0.00005, same direction in all three.

Pass rate was 91% injected against 80% baseline pooled, ahead in every
replication, but p=0.065, so treat it as directional and not established.

An earlier internal report quoted a 60% tool-call reduction plus input-token,
latency and output-token gains. **Those did not replicate and are withdrawn.**
The replication also found the eval's actors were fixed strings, so both arms
shared one durable workspace across campaigns and a later baseline opened onto
an earlier run's solved fixtures. Fixed with a per-replication actor tag. The
40% figure is the number measured after that fix, and it is the only one we
quote.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/872"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790760929&installation_model_id=19911&pr_number=872&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F872&signature=78fd516f358764a81c4e650566da37af514a185abf7e5b2354da80269d87a7f9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,8 +42,8 @@ BUDGET_WINDOW_MS=86400000
 # Extra root CA (PEM content) trusted for the Postgres connection, for providers that pin a private root. Optional; verification stays on. (core)
 # DATABASE_CA_CERT=
 
-# Procedural memory (Memorable). Off unless MEMORABLE is a true value; QM_MEMORABLE=0
-# forces it off regardless. MEMORABLE_BIN names the binary to spawn (split on spaces,
+# Procedural memory (Memorable). Off unless MEMORABLE is a true value; any false value of
+# QM_MEMORABLE forces it off regardless. Both are read once at startup. MEMORABLE_BIN names the binary to spawn (split on spaces,
 # so "npx memorable" works); a missing binary is a silent no-op.
 # See docs/procedural-memory.md.
 #MEMORABLE=1
--- a/.env.example
+++ b/.env.example
@@ -53,3 +53,8 @@ BUDGET_WINDOW_MS=86400000
 #MEMORABLE_API_URL=
 # Fallback key for scopes with no connected account of their own.
 #MEMORABLE_API_KEY=
+# The spawned CLI is given MEMORABLE_BACKEND=qm and MEMORABLE_DB_URL=DATABASE_URL
+# unless you set them here, so procedures land in this deployment's own Postgres
+# rather than a per-machine file. Set them only to override that.
+#MEMORABLE_BACKEND=qm
+#MEMORABLE_DB_URL=
--- a/.env.example
+++ b/.env.example
@@ -49,3 +49,7 @@ BUDGET_WINDOW_MS=86400000
 #MEMORABLE=1
 #QM_MEMORABLE=0
 #MEMORABLE_BIN=memorable
+# Where per-scope device sign-in is performed. Defaults to the public service.
+#MEMORABLE_API_URL=
+# Fallback key for scopes with no connected account of their own.
+#MEMORABLE_API_KEY=
--- a/.env.example
+++ b/.env.example
@@ -41,3 +41,11 @@ BUDGET_WINDOW_MS=86400000
 
 # Extra root CA (PEM content) trusted for the Postgres connection, for providers that pin a private root. Optional; verification stays on. (core)
 # DATABASE_CA_CERT=
+
+# Procedural memory (Memorable). Off unless MEMORABLE is a true value; QM_MEMORABLE=0
+# forces it off regardless. MEMORABLE_BIN names the binary to spawn (split on spaces,
+# so "npx memorable" works); a missing binary is a silent no-op.
+# See docs/procedural-memory.md.
+#MEMORABLE=1
+#QM_MEMORABLE=0
+#MEMORABLE_BIN=memorable
--- a/README.md
+++ b/README.md
@@ -173,52 +173,30 @@ messages, and screenshots for organization identifiers before it pushes. Nothing
 
 ## Procedural memory (Memorable)
 
-QM can remember _how_ it solved a task, not just what was said. With the
-[Memorable](https://github.com/NIkhil-cmd-cmd/memorable-qm) integration enabled, a finished
-session's tool-call trace is parsed — deterministically, no LLM involved — into a
-structured procedure (which files changed, which commands verified the work, in what
-order, with real exit codes) and stored in QM's own Postgres. When a later session starts
-on a similar task, a short pointer (~290 tokens) is recalled and injected into the
-prompt, telling the agent where the fix landed last time and authorizing it to skip
-re-diagnosis.
+Optional, off by default, additive. QM can record _how_ a task was done and replay that
+later, so a recurring job is not re-diagnosed from scratch every time. A prompt and the
+tool calls that followed it become one _memorable_: which files changed, which commands
+verified the work, in what order, with the real exit codes. A later prompt that one of
+them already answers gets a short pointer injected into the system prompt.
 
-- Off by default. Set `MEMORABLE=1` to enable the capture relay and recall injection;
-  `QM_MEMORABLE=0` is the kill-switch that wins over everything.
-- Consent is per scope and fail-closed: writes happen only for scopes explicitly set to
-  `read-write` via `memorable enable` (unset means deny).
-- Storage rides QM's own database (`memorable_procedures` / `memorable_mode` tables in
-  the same `DATABASE_URL` Postgres) — no second store.
-- Injected context is wrapped in a data-not-instructions envelope, control-character
-  stripped, and size-capped at write time.
-- A prompt that is several things at once can be answered with a **plan** instead of one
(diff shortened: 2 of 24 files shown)


16francej:
Merged as #894, which lands this as a pluggable memory provider. Thanks Nikhil!

PR #908, opened 2026-09-02
fix: reject empty non-surface Codex completions

m-mohamed:
A Codex completion with no text, files, approval, or explicit terminal outcome can be recorded as a successful empty reply. Reject that result on non-surface turns, while accepting files confirmed by the native attachment-staging tool.

Surface-tool turns remain under QM's existing delivery and addressed-message recovery logic. Approval, silence, and stopped outcomes retain their existing semantics. No automatic outbox harvesting or changing run identifiers are added to the system prompt.

Validation: 134 harness/agent-tool tests pass on Node 24.15, including empty replies, preserved surface recovery, zero versus positive staged attachments, and native tool approvals. TypeScript, affected ESLint, and independent review pass.

Live QA at this revision passed with real Codex: a non-surface harness turn returned the exact requested nonce. An isolated local QM instance with PostgreSQL also displayed the exact requested web reply. Empty-result rejection and the attachment/approval exceptions are exercised deterministically in the targeted tests. This change rejects empty non-surface completion; it does not establish delivery confirmation or require an explicit final_answer phase.

### DIFF
--- a/src/harness/agent-tools.ts
+++ b/src/harness/agent-tools.ts
@@ -76,6 +76,7 @@ export interface ToolContextRef {
   abortSignal?: AbortSignal;
   pollFire?: boolean;
   silentRequested?: boolean;
+  attachmentsStaged?: boolean;
 
   goal?: GoalRecord | null;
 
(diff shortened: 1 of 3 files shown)


m-mohamed:
Maintainer review requested. This fixes a production defect reproduced in a web project/group chat: completed tool work produced no durable terminal response, run/attempt identity was unavailable to the agent, and outbox evidence was absent from Files.

GitHub Actions are currently awaiting maintainer authorization for this outside-contributor PR. Local verification is recorded in the PR body: 175 affected tests passed, plus typecheck, lint, Prettier, and `git diff --check`. Maintainers may modify the branch.

PR #922, opened 2026-09-03
Add the agent37 sandbox backend

vishnukool:
Agent37 as a fifth agent-computer substrate, next to Sprites, smolmachines, Porter and AWS. The rationale is in `adrs/agent37-sandbox-backend.md`. If you would rather take the ADR alone and implement it yourselves, as CONTRIBUTING suggests, say so and I will trim the PR to that file.

What changes

- `SANDBOX_BACKEND=agent37` with `AGENT37_API_KEY`; optional `AGENT37_API_BASE_URL`, `AGENT37_NAME_PREFIX`, `AGENT37_TEMPLATE`, `AGENT37_CPUS`, `AGENT37_MEMORY_GB`, `AGENT37_DISK_GB`, `AGENT37_EGRESS_PROXY_URL`.
- `src/sandbox/agent37-sandbox.ts`, modeled on the smolmachines backend: one instance per scope, created on first use, deleted on destroy. Files ride the exec endpoint in base64 chunks, like the Porter backend. Commands past the 240 s sync ceiling run detached and are polled; output past 128 KiB is spooled and read back in 256 KiB chunks, under the API's 512 KiB cap. A command sent to a stopped or sleeping instance starts it and retries, including while a checkpoint is in flight.
- Auto sleep is off for now: Agent37 counts only traffic through the instance's routed port as activity, and qm drives the machine through exec alone. Once exec counts as activity on our side this becomes a one-line change.
- `agent37` added everywhere a backend name is enumerated: routing union, config, wiring, runtime secret schema, CLI secret spec, dev-instance script.

Tests

- `test/agent37-sandbox.test.ts` against `test/support/fake-agent37.ts`, an in-process fake of the Agent37 API, so CI needs no account.
- Verified locally: typecheck, eslint, oxlint, the new suite (22 tests), config, secret-schema conformance and drift, sandbox routing, and the CLI suite.


vishnukool:
Pushed one more commit: auto sleep is now on for the computers. Agent37's exec endpoint wakes a sleeping instance itself and counts every command as activity while it runs (shipped on our side today), so an idle computer sleeps and bills disk alone without ever being frozen mid-command. Tests and ADR updated to match.

Issue #845, opened 2026-08-29
claude-harness closes the prompt queue while background subagents are still running: every fan-out dies when the parent's turn settles, and the children report a phantom infrastructure outage


profd2004:
## Summary

`src/harness/claude-harness.ts` closes the SDK prompt queue as soon as the parent's turn produces a `result`, without checking whether any subagent is still running. The `Agent` tool starts a **background** task, so the parent routinely settles while its children are mid-task. Closing the queue ends the CLI session and takes the in-process MCP server (`createSdkMcpServer({ name: "qm", ... })`) down with it.

From that moment every subagent tool call returns **empty content with no error**. The call never reaches core at all — nothing is recorded in `run_activity`, nothing throws, nothing is logged.

## Why this is worth fixing beyond the hang

The failure mode is not "the fan-out stops". It is "the fan-out lies".

A subagent whose `execute`, `read`, `memory` and `background` calls all return empty, in sequence, with no error, has exactly one reasonable conclusion available to it: the machine is gone. So it writes that into its handoff, and the parent aggregates four such handoffs into a report of a platform outage. Verbatim, from four children on one run:

> every tool call — `execute`, `background`, `read`, `memory`, and even the out-of-band `computer: status` / `computer: restart` — began returning empty results with no output and no error. I probed roughly ten times across several minutes, including a machine restart, with no recovery. This is not something my commands caused; it's the sandbox transport itself.

> Group E is BLOCKED — not done. No commits were made. The worktree is untouched and clean. […] My sandbox went unresponsive partway through. Every tool call — `execute`, `read`, `background`, and both the out-of-band `computer: status` and […]

> The environment went down partway through. Everything below the line is verified; nothing was written, run, or committed — the worktree is exactly as I found it. […] After my first six tool calls succeeded, every subsequent command […]

> Tooling went down mid-task and did not recover. Reporting what stands and what does not.

Every one of those was false. On our side this drove two separate infrastructure investigations — node health, container runtime, CPU throttling, DNS, cgroup limits — before anyone thought to correlate the children's death against the *parent's* turn boundary. Two unrelated real bugs were found and fixed along the way, and neither changed the symptom, because neither was the cause.

A silent empty result is the worst possible shape for this. An error string, any error string, would have ended it in minutes.

## Root cause

`src/harness/claude-harness.ts` (935-line version), in the `for await (const message of sdkQuery)` loop:

```ts
// L701
if (pendingPrompts > 0) continue;
if (!signalsStopped) {
[...]

PR #945, opened 2026-09-04
claude harness: end the turn on the SDK's command lifecycle, not a result-per-prompt count

ReganBell:
## Problem

Users on the Claude harness were seeing `I hit an error and couldn't finish — Claude turn exceeded 2000s wall clock` (or whatever `TURN_WALL_CLOCK_SEC` is set to) on turns that had in fact already produced their reply. With no wall clock configured the same turns hung until the run reaper.

The trigger is a mid-turn steer. The Claude Agent SDK runs Claude Code as a child process with a streaming input queue. A user message pushed while a turn is in flight is **folded into that running turn** at the next tool boundary and shares its single `result` message. The harness assumed one `result` per prompt it pushed (`pendingPrompts`), so after a folded steer it waited for a second `result` that never came, idled until the wall clock fired, and threw the wall-clock error in place of the reply it had already streamed.

Verified against the real SDK (0.3.211 / CLI 2.1.259):

- steer pushed during a tool call → folded into the turn, **one** `result`, harness hung
- steer pushed after the model's last message → its own turn, **two** `result`s, harness fine

## Fix

The SDK emits `command_lifecycle` events (`queued` → `started` → `completed`/`cancelled`) for every uuid-stamped inbound user message and advertises this as the `msg_lifecycle_v1` capability on `system/init`. The harness now:

- stamps every pushed user message with a `uuid`
- tracks open command uuids in a set and settles one on a terminal lifecycle state
- closes its prompt queue (after stopping the signal poll) when the set empties, whether a command ran as its own turn or was folded into an earlier one
- falls back to the old one-result-per-prompt accounting, through the same settle path, when `init` does not advertise the capability (older CLIs)

## Tests

- New: a steer folded into the running turn ends the turn on the lifecycle signal. Times out at 5s on the previous code, passes now.
- New: a steer queued behind the model's last reply still gets its own turn and the harness waits for it.
- Existing steer tests exercise the legacy (no-capability) path unchanged.
- Live: ran the real harness against the real SDK with a steer sent during an `Agent` subagent call; the steer was absorbed into the reply and the turn returned in ~15s.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/945"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791137314&installation_model_id=19911&pr_number=945&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F945&signature=29e4cfec957208fce30b3ee9f590b48e8fb7a3197f8901817897e0889738678e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/harness/claude-harness.ts
+++ b/src/harness/claude-harness.ts
@@ -1,4 +1,4 @@
-import { randomBytes } from "node:crypto";
+import { randomBytes, randomUUID } from "node:crypto";
 import { spawn } from "node:child_process";
 import { chownSync, mkdtempSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
@@ -181,11 +181,12 @@ class MessageQueue implements AsyncIterable<SDKUserMessage> {
   private readonly waiters: Array<(value: IteratorResult<SDKUserMessage>) => void> = [];
   private ended = false;
 
-  push(value: SDKUserMessage): void {
-    if (this.ended) return;
+  push(value: SDKUserMessage): boolean {
+    if (this.ended) return false;
     const waiter = this.waiters.shift();
     if (waiter) waiter({ value, done: false });
     else this.values.push(value);
+    return true;
   }
 
   close(): void {
@@ -277,9 +278,16 @@ function promptText(turn: HarnessTurnInput): string {
   return [replay, prior, turn.input, turn.environment].filter((value) => value?.trim()).join("\n\n");
 }
 
+function commandLifecycle(message: SDKMessage): { commandUuid: string; settled: boolean } | null {
+  const raw = message as { type?: string; command_uuid?: string; state?: string };
+  if (raw.type !== "command_lifecycle" || typeof raw.command_uuid !== "string") return null;
+  return { commandUuid: raw.command_uuid, settled: raw.state !== "queued" && raw.state !== "started" };
+}
+
 function userMessage(text: string, images: HarnessTurnInput["images"] = []): SDKUserMessage {
   return {
     type: "user",
+    uuid: randomUUID(),
     message: {
       role: "user",
       content: [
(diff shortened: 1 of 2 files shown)


punya10:
Independent confirmation from a self-hosted v0.1.9 (SDK 0.3.211, bundled Claude Code 2.1.211). Same root cause, same symptom: every run that received a mid-turn steer stayed `running` with an idle model until the reaper.

Numbers from one run (`77f5fc8e`): 1 initial prompt + 3 steers = 4 pending, the model coalesced them into 2 results, so `pendingPrompts` ended at 2 and the queue never closed.

We have been running a different fix in production since 2026-09-21 that does not depend on `msg_lifecycle_v1`: an input ledger on the prompt queue (delivered / consumed / deferred), with the result handler closing the queue once everything delivered has been consumed at the queue iterator, instead of counting results. A steer that lands after close is retained rather than dropped. A run with two mid-turn steers coalesced into one result now finishes on its own (63 s in our test). Happy to share the diff, but this PR's approach works too. For what it is worth, the bundled 2.1.211 binary already contains the `msg_lifecycle_v1` and `command_lifecycle` strings, so the lifecycle path here would apply to the shipped CLI, not only the fallback.

One request: #845 is the mirror-image bug at the same `queue.close()` site (closing while background subagents are still `in_progress`). Whatever lands here should probably consult `taskStates` as well, otherwise the fan-out case stays broken.

This PR has had no review in 18 days and `claude-harness.ts` has moved ten commits since, so it likely needs a rebase. It is blocking real use of the Claude harness with steering.

PR #959, opened 2026-09-06
fix(web-ui): stop animating programmatic transcript scrolls

time-attack:
## Problem

When a run ended with a pending approval, the web-ui transcript visibly scrolled from the top of the conversation all the way to the bottom as an animation.

The transcript scroller (`.chat-scroll`) had `scroll-behavior: smooth`. That property only affects programmatic scrolls (assigning `scrollTop`), never wheel or trackpad input. Every redraw that re-sticks the transcript to the bottom therefore animated, and when the scroller had been rebuilt at the top the animation covered the whole conversation.

## Change

- Remove `scroll-behavior: smooth` from `.chat-scroll`.
- Delete the three call sites that flipped `style.scrollBehavior` to `auto` and back to dodge the animation (forced scroll-to-bottom, and both pagination anchor adjustments). They are now plain `scrollTop` writes.
- The layout test now asserts no smooth scrolling in the transcript CSS and no `scrollBehavior` juggling in `chat.ts`, and still asserts both pagination anchors exist.

Net: 5 insertions, 20 deletions.

## Demo

Behavior-only change: the after state renders pixel-identical to before, the difference is that scroll-to-bottom is now instant instead of animated. No screenshot can show it, so none attached. To see it: open a long conversation, trigger a run that ends in an approval, and observe the transcript snaps to the pending approval instead of animating from the top.

## Verification

- `plugins/web-ui`: `layout-thrash`, `tab-switch-scroll-snap`, `pane-approval-overflow` tests pass.
- `plugins/web-ui` typecheck passes; eslint clean on changed files.

## Not in this PR

The screenshot that prompted this also asked for an "allow always" option on security-screen quarantine releases. Those are deliberately once-only in the orchestrator (server-side guard plus tests). The existing knob is the `security-posture` admin resource (`dangerous` disables content screening). Left unchanged here.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/959"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791245311&installation_model_id=19911&pr_number=959&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F959&signature=6adead0b6678746c5b3a2f2c479f381309d0472ec63d85b9493745f4af3e8d08"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/web-ui/src/chat.ts
+++ b/plugins/web-ui/src/chat.ts
@@ -900,10 +900,7 @@ export function createChatSurface(
                               requestAnimationFrame(() => {
                                 const scrollerNow = container?.querySelector<HTMLElement>(".chat-scroll");
                                 if (!scrollerNow) return;
-                                const prev = scrollerNow.style.scrollBehavior;
-                                scrollerNow.style.scrollBehavior = "auto";
                                 scrollerNow.scrollTop = priorTop + (scrollerNow.scrollHeight - priorHeight);
-                                scrollerNow.style.scrollBehavior = prev;
                               });
                             } catch {
                               btn.disabled = false;
@@ -1002,10 +999,7 @@ export function createChatSurface(
       requestAnimationFrame(() => {
         const scrollerNow = chatState.host?.querySelector<HTMLElement>(".chat-scroll");
         if (!scrollerNow) return;
-        const prev = scrollerNow.style.scrollBehavior;
-        scrollerNow.style.scrollBehavior = "auto";
         scrollerNow.scrollTop = priorTop + (scrollerNow.scrollHeight - priorHeight);
-        scrollerNow.style.scrollBehavior = prev;
       });
     } catch {
       void 0;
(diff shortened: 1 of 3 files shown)


time-attack:
Independent review pass (code-review, medium, eight angles) on the first commit found no correctness issues. Two cleanups it raised are applied in the follow-up commit:

- The two pagination anchor-restore blocks were verbatim duplicates once the smooth-scroll workaround was gone. They now share one `holdScrollAnchor` helper.
- The layout test no longer pins the duplication at exactly two copies. It asserts both loaders route through the helper.

Confirmed root cause from the altitude angle: `mountContinuable` and `mountReadOnly` build a fresh host at scrollTop 0 and snap to the bottom a frame later, which `scroll-behavior: smooth` turned into a full-transcript animation. Removing the CSS rule fixes every current and future programmatic scroll on that element in one place.

Left as pre-existing and out of scope: `scrollTranscript` has no pending-rAF coalescing guard, and the two mount paths duplicate the build-then-snap shape.

PR #964, opened 2026-09-07
feat(web-ui): open image attachments in a gallery viewer

time-attack:
## What

Clicking an image attachment in a chat message used to open the raw artifact URL in a new tab. It now opens the image in a minimal viewer, ChatGPT-style: the page dims, the image is centred, and a Download link and a Close button sit in the top-right corner. Nothing else.

- Click any image in a message to open it. The viewer is a native `<dialog>` opened with `showModal()`, mounted once on `document.body`.
- `Esc`, the Close button, or a click on the dark backdrop closes it and returns focus to the tile that opened it. Escape is handled by the existing `trapDialogFocus` helper, so it no longer reaches the page-level handlers (split-pane un-maximise, sidebar selection clear).
- When the message has more than one image, `←`/`→` and two small prev/next arrows step between them (wrapping). With a single image there are no arrows and the keys are a no-op.
- Download links to the stored artifact when there is one, otherwise to the inline data URL.

Implementation: one new module `plugins/web-ui/src/lightbox.ts` (about 100 lines), `chat.ts` routes user attachments and delivered files through one `messageFiles` helper that builds the per-message gallery and renders image tiles as `<button>`s, and the duplicated data-URL construction moved to `inlineImageSrc` in `ui.ts`.

**Behaviour change to note:** transcript image tiles are now `<button>`s instead of `<a target="_blank">`, so cmd/middle-click to a new tab no longer works on the tile itself. Download inside the viewer covers getting at the file.

## Changes since review

The first cut was heavyweight (header bar with name, size and counter, filmstrip, zoom-to-natural-size, touch swipe, live-region announcer, composer attachment-strip previews, square cover-tile grid). All of that is gone; the viewer is now image + Close + Download, plus arrows only for multi-image messages. Diff went from +788/-46 across 8 files to +342/-29 across 6 files; CSS is about 70 lines; the test file covers open/close, navigation and wrap, single-image mode, Escape containment with focus return, and backdrop-vs-image click.

## Demo

Viewer open on image 2 of 3 (static demo page against the real `shell.css`; images are placeholder gradients):

viewer

## Verification

- `plugins/web-ui`: `npm run typecheck` (all three tsconfigs) clean; eslint and prettier clean on the changed files.
- `test/lightbox.test.ts` (jsdom via vite `ssrLoadModule`) and `test/renderable-image-source.test.ts`: 6/6 pass.
- Two independent fresh-context reviews of the diff (Escape containment, focus return, backdrop vs image click, navigation bounds, chat.ts fallbacks for SVG / non-image / inline-only attachments). One real bug found and fixed: clicking the image moved focus to `body`, so a following Escape bypassed the dialog's keydown handler and reached page-level handlers before the native close ran; `tabindex="-1"` on the dialog keeps focus inside it. Nits taken: focus falls back to the composer input when the opening tile was re-rendered away, the dialog label announces "Image n of m", phone-width padding uses `clamp()`, and `inlineImageSrc` takes a required mime since both callers already check it. Full list in the PR comment.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/964"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791343415&installation_model_id=19911&pr_number=964&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F964&signature=d0158dfb40667bf0e2abdc452f6d50ba7782aa54bd37696e77a1470e5a403425"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/web-ui/README.md
+++ b/plugins/web-ui/README.md
@@ -49,7 +49,9 @@ and `CORE_SIGNING_SECRET` (same value as the core when source-auth is enabled).
   transcript, bottom composer, inline **model selector** (the models core reports as
   serviceable for the approved harnesses),
   explicit **effort selector** (`low|medium|high|xhigh|max|ultracode|auto`), **Fast mode**
-  toggle, attachments, streaming partials, and a theme toggle.
+  toggle, attachments (image attachments open in a full-screen viewer with arrow-key,
+  swipe, and filmstrip navigation across every image in the message), streaming partials,
+  and a theme toggle.
   The UI drives Pi's `Agent` with a custom `streamFn` (`src/core-bridge.ts`) instead of
   mounting Pi's stock `AgentInterface`.
 - **Slash-command skill picker** — type `/` at the start of the composer for a Codex-style
--- a/plugins/web-ui/README.md
+++ b/plugins/web-ui/README.md
@@ -49,9 +49,7 @@ and `CORE_SIGNING_SECRET` (same value as the core when source-auth is enabled).
   transcript, bottom composer, inline **model selector** (the models core reports as
   serviceable for the approved harnesses),
   explicit **effort selector** (`low|medium|high|xhigh|max|ultracode|auto`), **Fast mode**
-  toggle, attachments (image attachments open in a full-screen viewer with arrow-key,
-  swipe, and filmstrip navigation across every image in the message), streaming partials,
-  and a theme toggle.
+  toggle, attachments, streaming partials, and a theme toggle.
   The UI drives Pi's `Agent` with a custom `streamFn` (`src/core-bridge.ts`) instead of
   mounting Pi's stock `AgentInterface`.
 - **Slash-command skill picker** — type `/` at the start of the composer for a Codex-style
--- a/plugins/web-ui/src/chat.ts
+++ b/plugins/web-ui/src/chat.ts
@@ -2292,7 +2292,7 @@ export function createChatSurface(
     if (!browserRenderableImage(a.mimeType)) return undefined;
     const href = artifactHref(a.artifactId);
     const src = href ?? (a.content ? inlineImageSrc(a.mimeType, a.content) : undefined);
-    return src ? { src, name: a.fileName, size: a.size, href } : undefined;
+    return src ? { src, name: a.fileName, href } : undefined;
   }
 
   function userAttachmentChip(a: UserAttachmentView): TemplateResult {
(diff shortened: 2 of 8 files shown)


time-attack:
Reworked per the "heavyweight" feedback. The viewer is now ChatGPT-style: dim the page, centre the image, Download and Close in the top-right corner, and two small prev/next arrows only when the message has more than one image (`←`/`→` do the same). Esc, backdrop click, or Close dismisses and returns focus to the tile.

Cut: header bar (file name, size, `n / m` counter), filmstrip thumbnails, zoom-to-natural-size, touch swipe, live-region announcer, "Open original" link, composer attachment-strip previews, the square cover-tile grid for multi-image messages, and the README paragraph. `composer.ts` and `README.md` are back to their pre-PR state.

Kept: the `messageFiles` consolidation in `chat.ts` (one path for user attachments and delivered files) and `inlineImageSrc` in `ui.ts`, which together remove the duplicated data-URL and href logic.

Diff: +788/-46 across 8 files, now +342/-29 across 6 (lightbox.ts ~100 lines, CSS ~70, tests trimmed to 5 covering open/close, navigation and wrap, single-image, Escape containment with focus return, backdrop vs image click). Fresh screenshot in the body.

Two independent fresh-context reviews of the new diff. Fixed: clicking the image moved focus to `body`, so a following Escape skipped the dialog's keydown handler and reached page-level handlers before the native close (`tabindex="-1"` on the dialog). Also taken: focus falls back to the composer input if the opening tile was re-rendered away, the dialog label reads "Image n of m", `clamp()` padding on phone widths, `inlineImageSrc` mime typed as required. Not taken: `autofocus` instead of the explicit `.focus()` (jsdom can't exercise the dialog focusing steps, so the test would lose its assertion), and a mousedown-on-image / mouseup-on-backdrop slip closing the viewer (standard lightbox behaviour). One pre-existing sibling noted for a separate change: `.project-dialog` in `contexts.ts` relies on native Escape without `trapDialogFocus`, so the same body-focus leak applies there independently of this PR.

PR #976, opened 2026-09-08
web-ui: a UI pass over the chat, inbox, settings, and sidebar surfaces

evebouf:
Twenty-four small changes, all one- or two-file UI work. One line each.

**Chat page**

- Transcript edges clip instead of fading behind a gradient and a blur.
- The pinned prompt paints the background and covers what scrolls under it, instead of fading it.
- Collapsed plain-text code fences clip their overflow instead of fading it.
- A long pinned prompt gets **Show more** instead of a second scrollbar inside the transcript.
- Message hover actions are hittable: the copy button goes from an 18px square to 24px, gap 6px to 8px.

**Inbox item page**

- The page scrolls instead of squeezing the thread into the viewport; the draft may fall below the fold.
- The thread caps at 480px and offers **Show the whole thread** when there's more.
- An email draft folds **To**, **Cc**, and **Subject** away, collapsed by default, behind a line naming the recipient and subject.
- The folded fields get 8px between them, which the first cut lost by putting the gap on the `details` element.
- The header anchors to the top of its row instead of floating in the middle of one the assistant stretched.
- The assistant ends at the pane's content box, so its own height stops scrolling the page.
- The assistant tracks the window between a 320px floor and a 1100px ceiling.
- The assistant's size observer is replaced on remount rather than stacked.
- The empty assistant offers three starting prompts: shorter, friendlier, drop the salutations.
- The inbox shows the same working wave as the rest of the app.
- The sync control uses the app's tooltip and the shared icon size.
- The inbox list drops its subtitle.

**Settings page**

- The theme picker sits right like every other row's control, instead of wrapping to its own line.
- **Import theme file** is a small link rather than a bordered button competing with the theme choice.

**Sidebar and filters**

- Archive and the per-chat overflow button get tooltips; they had an `aria-label` and nothing visible.
- Count badges use tabular figures so a badge stops reflowing its row as digits change.
- A nav entry always lands on its own index.
- The context and surface filters become one control.
- Filter menus open below their button rather than over the header above, and drop the header that repeated the button.
- List filter captions stop bolding the dropdown they label, at five call sites.

**Everywhere**

- Dark mode gets a `--scrim` token, so a modal dims the page instead of washing it out with near-white.
- Dark mode defines `--accent-foreground` instead of letting three rules fall back to `#fff`.

web-ui 973/973, typecheck, oxlint, knip, prettier clean. Exercised against a live dev instance at 620px, 880px, and full height, in both themes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ReganBell (approved):
send!

PR #980, opened 2026-09-08
fix: accept custom models the web picker already offers

### LINKED ISSUE #952: Web turn validation falls back to DEFAULT_WEBUI_MODEL_IDS and refuses custom models the picker offers (surfaces as bare HTTP 403)
**Version:** CLI 0.1.7 (core image from the 0.1.7 manifest). Harness: `pi`.

## What happened

With a custom provider registered (`PUT /v1/admin/custom-providers/...`), the web-UI picker lists the custom model (surface.ts builds `webuiModels` from the live catalog, which includes custom models). Selecting it and sending a message fails with a bare **`HTTP 403`** in the web UI.

## Why

`validateWebTurnModelOptions()` (src/core/turn-options.ts:47-58) is called from `app.turn()` with `enabledWebuiModels`. When the org has **no** web-UI allowlist configured and no OpenRouter key, `enabledWebuiModels` is `null`, so the function falls back to `DEFAULT_WEBUI_MODEL_IDS` — the hardcoded registry list — which never contains custom models. The turn is refused with "that model is not enabled for the web UI"; `postTurn` maps `refused` to 403 (src/api/routes/turns.ts); and the web-UI client shows `o?.message ?? o?.error ?? \`HTTP ${status}\``, but the refusal body carries `reason`, not `message`/`error`, so the operator only sees `HTTP 403` with no explanation.

So the picker and the validator disagree: the picker offers the model, the validator refuses it. Workaround: configure an explicit allowlist in Governance → Allowed models that includes the custom model.

## Suggested fix

- In `validateWebTurnModelOptions`, treat the fallback set as `DEFAULT_WEBUI_MODEL_IDS ∪ customModelCatalog().map(m => m.id)` (same source the picker uses), or pass the picker's computed `allowed` list through.
- Surface `reason` from a `refused` turn in the web UI instead of the bare status.

## Repro

1. Fresh org, no "Allowed models" configured, no OpenRouter key.
2. Register any custom provider with one model via Admin → Onboarding → Custom providers.
3. Web chat → pick the custom model → send a message. Expected: turn runs. Actual: `HTTP 403`; nothing in `error_events`.

siddiqueirshad:
## Why

With no web-UI allowlist and no OpenRouter key, `GET /v1/surface-config` still lists admin-registered custom models (from `builtInModelCatalog()`). Sending a turn on one of those models was refused: `validateWebTurnModelOptions` fell back to `DEFAULT_WEBUI_MODEL_IDS`, which never includes custom ids. `postTurn` maps that refusal to HTTP 403.

That is the documented custom-provider path. The picker and the gate disagreed.

Fixes #952

## What

- Default web-UI model ids now include the live custom catalog, matching overlays.
- When no allowlist is configured, web-turn validation uses the same catalog as the picker (`builtInModelCatalog`, or the OpenRouter catalog when that key is present) instead of leaving the allowlist `null`.
- An explicit Governance allowlist still refuses custom models that are not on it.

## Tests

- Unit: fallback allowlist includes a registered custom model; an explicit allowlist still refuses it.
- Integration: register a custom provider with no OpenRouter/OpenAI env key, confirm the picker lists it, a web turn queues, then an allowlist that omits it refuses with `that model is not enabled for the web UI`.
- Focused suites: `test/turn-options.test.ts`, `test/custom-providers.test.ts`, `test/custom-provider-route.test.ts`, `test/model-registry.test.ts` — 33 passed.

Made with [Cursor](https://cursor.com)

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/980"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791470342&installation_model_id=19911&pr_number=980&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F980&signature=21840cf2806e658b541c98372bbee57580b2294645f71ccda196e0ad5a7da86f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/api/app-turn.ts
+++ b/src/api/app-turn.ts
@@ -19,7 +19,7 @@ import {
   modelUnavailableReason,
   resolveModel,
 } from "../model/pi-models.ts";
-import { selectableCatalogForHarness, selectableModelCatalog } from "../model/model-catalog.ts";
+import { builtInModelCatalog, selectableCatalogForHarness, selectableModelCatalog } from "../model/model-catalog.ts";
 import { resolveRuntimeChoiceDurable } from "../harness/harness-router.ts";
 import { swallow, swallowAs } from "../util/errors.ts";
 import { sleep } from "../util/async.ts";
@@ -204,13 +204,13 @@ export function createTurnMethods(
             enabledWebuiModels = configuredWebuiModels.length
               ? [...new Set([...configuredWebuiModels, orgRuntime.modelId])]
               : [];
-          } else if (providers?.openrouter) {
+          } else {
+            const catalog = providers?.openrouter
+              ? await selectableModelCatalog(deps.modelCredentialFetch)
+              : builtInModelCatalog();
             enabledWebuiModels = [
               ...new Set([
-                ...selectableCatalogForHarness(
-                  await selectableModelCatalog(deps.modelCredentialFetch),
-                  runtime.harnessId,
-                )
+                ...selectableCatalogForHarness(catalog, runtime.harnessId)
                   .filter((model) => modelOfferedInWebui(model.id))
                   .map((model) => model.id),
                 ...(orgRuntime.harnessId === runtime.harnessId ? [orgRuntime.modelId] : []),
--- a/src/model/pi-models.ts
+++ b/src/model/pi-models.ts
@@ -2,7 +2,7 @@ import { getBuiltinModel } from "@earendil-works/pi-ai/providers/all";
 import type { Api, Model } from "@earendil-works/pi-ai";
 import { parseModelOverlay, type ModelOverlay } from "./model-overlay.ts";
 import { providerBaseUrl } from "./provider-endpoints.ts";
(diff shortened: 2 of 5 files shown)


siddiqueirshad:
Hi @16francej, when you get a chance, could you please take a look at this PR? Thanks!

PR #1017, opened 2026-09-09
Fix ChatGPT subscription logins expiring silently after the first access token

brAperia2025:
## Problem

A ChatGPT-subscription connection made through the device-code login works until its first
access token expires, then every turn on that connection fails hard. The failure is silent
until then, so it looks like the feature broke days after it was set up.

Mechanism:

- `tokensFromAuthJson` (the device-login harvest) returns the tokens without an
  `expiresAt`, so the keychain record is stored with no expiry.
- The keychain's central refresh (`connectorTokenForRecord`) only engages when the stored
  record carries `expiresAt`; with none, the original access token is handed out forever.
- The Codex child jail deliberately receives no usable refresh token
  (`refresh_token: ""`), so once the access token ages out the CLI's own refresh attempt
  fails with "Failed to refresh token: … got an empty string instead" and the turn dies.

Reconnecting only resets the clock: the fresh harvest also stores no expiry.

## Fix

- The login harvest now carries the access token's own `exp` claim (via
  `codexOAuthAccessTokenExpiresAt`) into the stored tokens, so the central refresh engages
  before expiry and rotated tokens persist normally from then on.
- `connectorTokenForRecord` falls back to the access token's `exp` claim when the record
  has no stored expiry. This heals records created before this fix on their next use: the
  stale token triggers a central refresh, and the refreshed record carries a real
  `expiresAt`. Opaque (non-JWT) access tokens — Google, Slack, and similar connectors —
  read no expiry from the fallback and behave exactly as before.

## Tests

- `test/codex-device-login.test.ts`: the fake login now writes a JWT access token and the
  harvest assertion pins `expiresAt` to its `exp` claim.
- `test/keychain.test.ts`: a record stored without `expiresAt` whose JWT access token has
  expired is refreshed on materialize; an opaque token without `expiresAt` is handed out
  unchanged with no refresh attempt.

`npm run typecheck` clean; the new keychain case and the full CLI-visible behavior are
covered by the suites above.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/1017"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791560204&installation_model_id=19911&pr_number=1017&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F1017&signature=45b633839d6f5a64e072370b9df03071e0600726169e7101d7071babf556a152"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/credentials/keychain.ts
+++ b/src/credentials/keychain.ts
@@ -459,6 +459,18 @@ function expired(rec: { expiresAt?: number }, now: number): boolean {
   return typeof rec.expiresAt === "number" && rec.expiresAt < now;
 }
 
+/** Epoch ms of a JWT access token's exp claim; undefined for opaque tokens. */
+function jwtExpiryMs(token: string | null): number | undefined {
+  if (!token || token.split(".").length !== 3) return undefined;
+  try {
+    const payload = JSON.parse(Buffer.from(token.split(".")[1] ?? "", "base64url").toString("utf8")) as unknown;
+    const exp = (payload as { exp?: unknown } | null)?.exp;
+    return typeof exp === "number" ? exp * 1000 : undefined;
+  } catch {
+    return undefined;
+  }
+}
+
 function credExpired(rec: { kind: CredentialKind; expiresAt?: number }, now: number): boolean {
   return rec.kind !== "file" && expired(rec, now);
 }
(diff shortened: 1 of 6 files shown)


brAperia2025:
Pushed a second commit covering a sibling defect with the same user-visible failure: the CLI reads an auth.json without `last_refresh` as needing an immediate token refresh, and the child holds only the empty refresh-token placeholder — so per-user turns failed at startup with "Failed to refresh token: … got an empty string instead" even with a freshly minted access token (verified against codex 0.144.5: the identical auth.json succeeds with the stamp and fails without it). `childCodexAuthFromDerived` now stamps the materialization time, and `childCodexOAuthAuth` stamps a fallback when the stored record never carried one. Both fixes are needed for subscription-served turns to survive past the first token lifetime.

PR #1040, opened 2026-09-10
Fix Slack bot-ID mentions without racing app mentions

### LINKED ISSUE #630: Bot-id-form channel mentions (`<@B…>`) aren't recognized: top-level ones are silently dropped, in-thread ones misfire as ambient

time-attack:
## Summary

- recognize Slack's bot-ID mention form alongside the bot-user-ID form in shared mention parsing, stripping, mirroring, and history rendering
- route validated human-authored bot-ID-only messages through the existing addressed-turn path without racing normal `app_mention` delivery
- establish thread stake only after accepted fallback intake, keeping rejected/failed/peer-bot cases fail-closed
- preserve existing local/durable dedupe keys

Fixes #630.

## Test plan

- `node --experimental-test-module-mocks --test --test-concurrency=2 test/slack-*.test.ts test/loop-source-send.test.ts` (543 passed)
- `npm run typecheck`
- `npm run lint:knip`
- ESLint and Prettier on the 9 changed files

Coverage is two integration cases in the existing `slack-index` harness (one addressed turn across message/app_mention/redelivery fan-out with mirror and history rendering; stake seeded only after core accepts, never for bot_id/bot_profile/bot_message/is_bot/failed-lookup authors) plus one unit case each for `mentionsBot`/`threadHasBotStake` and `stripMention`/`resolveMentionsInText`/`neutralizeMentions`. Each assertion was mutation-checked against a revert of the piece of the source change it covers. Live Slack client/event fan-out and real PostgreSQL multi-process replay remain follow-up validation.

## Changes since review

Trimmed the test additions from +920 to +93 lines: dropped the standalone `slack-bot-id-mentions` suite and the redelivery-dedup, loop-source-send, and event-order/retry permutation cases, which duplicated the integration harness or only asserted on log lines. Removed the rate-limited "bot-id mention accepted" diagnostic from `turn-handler.ts`, since only those deleted tests exercised it; `acceptBotIdFallback` now only marks thread stake.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/1040"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791592957&installation_model_id=19911&pr_number=1040&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F1040&signature=d7400d4c470d19b51a20e9cd7f0b89dfff3b265905d8a1cd65709744937f4d50"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


time-attack:
Trimmed per review: the PR is now +159/-28 (was +995/-28). Test additions went from +920 to +93 lines across three existing files; the standalone `slack-bot-id-mentions` suite and the redelivery-dedup / loop-source-send / event-order permutation cases are gone. Each remaining assertion was mutation-checked against a revert of the source piece it covers. Also removed the rate-limited "bot-id mention accepted" log from `turn-handler.ts`, which only those deleted tests exercised. Slack suite 543/543, typecheck, knip, ESLint, and Prettier all clean.

PR #1043, opened 2026-09-10
Make admin runtime editing scope-aware

### LINKED ISSUE #589: Admin UI only offers the harness/model control at org scope, though the resource accepts any scope

time-attack:
## Summary

- move the runtime override card from org-wide Models to scoped Governance and show the exact target being edited
- preserve scope across refresh and browser history while retaining stale-load, stale-write, and in-flight guards
- disable retained-but-unapproved harnesses and retained-but-unserviceable models without losing reset or valid-choice recovery
- expose durable approval and server-computed availability through the existing scope response, keeping authorization and validation unchanged

Fixes #589.

## Test plan

- admin UI/proxy/auth suites: 144 passed
- core authorization/resolution/catalog suites: 113 passed
- full-page Chromium suite against a local synthetic API: 6 passed
- independent manual browser QA: scoped target, refresh, Models isolation, Back restoration, and successful scoped runtime PUT
- root/admin typechecks, repository ESLint, Oxlint, Knip, targeted Prettier, and `git diff --check`

Coverage includes channel, project, personal, team, and organization scopes; inheritance reset; stale approvals; alternative/custom/empty catalogs; and authorization negatives.

## Before / after

Light theme, rendered against a synthetic catalog (`pi`/`codex`/`claude` harnesses, Claude and GPT models) with the shell served from each branch's `plugins/admin/public/index.html`.

| Before (`main`) | After (this branch) |
| --- | --- |
| Models view with the org-wide runtime card | Models view without the runtime card |
| Models view: the runtime card sits on the org-wide view with no indication of which scope it edits. | Models view: the runtime card is gone; only the org-wide cards remain. |
| No runtime card on Governance for non-org scopes; per-channel and per-project runtime could not be edited. | Governance view for channel:C0AB12CD3 with the scoped runtime card |
| | Governance for `channel:C0AB12CD3`: the card states the exact target and offers "Use organization default" to fall back to inheritance. |
| Retained-but-unapproved harnesses and unserviceable models rendered as ordinary selectable options. | Scoped runtime card with an unapproved harness and unavailable model disabled |
| | Retained `claude` harness (not approved) and its retained model render as "(unavailable)" and the card asks for an available choice before Apply is enabled. |
| | Harness menu with the unapproved harness disabled |
| | Harness menu open: the retained unapproved harness is disabled while approved harnesses stay selectable. |

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -13,6 +13,7 @@
       "entry": [
         "src/migrate-main.ts",
         "src/sandbox/local-sandbox.ts",
+        "scripts/admin-runtime.browser.mjs",
         "scripts/aws-deploy-smoke.ts",
         "scripts/aws-litestream-smoke.ts",
         "scripts/backfill-search-index.ts",
--- a/plugins/admin/public/index.html
+++ b/plugins/admin/public/index.html
@@ -5502,24 +5502,33 @@
                   <h2>Models &amp; browsing</h2>
                   <p>The models, browser runtime, and organization context available on future turns.</p>
                 </div>
-                <section class="card sv-models setting-row hidden" id="card-base-model">
+                <section class="card sv-governance setting-row hidden" id="card-base-model">
                   <div class="head">
                     <h2>Default runtime</h2>
+                    <p id="base-runtime-target"></p>
                   </div>
                   <div class="body">
-                    <label for="base-harness">Harness</label>
-                    <select id="base-harness" style="max-width: 360px"></select>
-                    <label for="base-model">Model</label>
-                    <select id="base-model" style="max-width: 360px"></select>
-                    <label for="base-effort">Reasoning level</label>
-                    <select id="base-effort" style="max-width: 360px"></select>
-                    <label class="setting-toggle" id="base-fast-mode-control" style="display: none; margin-top: 10px">
-                      <input type="checkbox" id="base-fast-mode" />
+                    <label class="setting-toggle hidden" id="base-inherit-control">
+                      <input type="checkbox" id="base-inherit" />
                       <span class="setting-switch" aria-hidden="true"></span>
-                      <span class="setting-copy"
-                        ><strong>Fast mode</strong><small>Only available for supported models.</small></span
-                      >
+                      <span class="setting-copy"><strong>Use organization default</strong></span>
                     </label>
+                    <p class="hint" id="base-runtime-note" aria-live="polite"></p>
+                    <div id="base-runtime-fields">
+                      <label for="base-harness">Harness</label>
+                      <select id="base-harness" style="max-width: 360px"></select>
+                      <label for="base-model">Model</label>
+                      <select id="base-model" style="max-width: 360px"></select>
+                      <label for="base-effort">Reasoning level</label>
+                      <select id="base-effort" style="max-width: 360px"></select>
+                      <label class="setting-toggle" id="base-fast-mode-control" style="display: none; margin-top: 10px">
+                        <input type="checkbox" id="base-fast-mode" />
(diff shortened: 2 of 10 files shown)


time-attack:
Added before/after screenshots to the description in response to review: the runtime card on the org-wide Models view on main, the Models view without it on this branch, the scoped card on Governance for a channel showing its target, and the disabled states for a retained unapproved harness and unavailable model.

PR #1056, opened 2026-09-10
web-ui: approval card with a pager

time-attack:
## Summary

The composer's approval panel becomes the reference card — title and reason
badge, summary and purpose, the command in a mono block, a "Show full command"
disclosure — showing one pending approval at a time with a ‹ 1/N › pager when
several are queued, ghost Deny / Allow once / Allow always pills and a primary
Allow for session. The transcript marker is the compact version. The pager
index resets with the conversation.

## Demo

A real pending approval (a pipe-to-shell command): the card with its reason badge, "Triggered by" snippet and the Deny / Allow pills.

## Verification

Typecheck, the full `plugins/web-ui` suite, eslint and prettier pass at this level of the stack (each level is verified on its own, on top of the previous one). Live QA of the finished stack ran in a browser-only `dev-instance` against a seeded conversation (tool run, markdown table, TypeScript and diff fences, a pending approval, ⌘K, list pages, settings), in light and dark and at phone width.

## Stack

Merge in order; each PR is based on the one above it and retargets automatically when its base merges.

| # | PR | Change |
| --- | --- | --- |
| 1 | #1053 | Beautiful UI foundation — tokens, fonts, styles scaffold |
| 2 | #1054 | loading state and the live thinking trace |
| 3 | #1055 | chat rows, session header and empty state |
| 4 | #1056 (this) | approval card with a pager |
| 5 | #1057 | tool chips and code blocks |
| 6 | #1058 | records tables and filter lists |
| 7 | #1059 | search palette |
| 8 | #1060 | settings as inspector cards |
| 9 | #1061 | selection actions toolbar |
| 10 | #1062 | recommendation card |

PR #1054, opened 2026-09-10
web-ui: loading state and the live thinking trace

time-attack:
## Summary

Loading: a 3×3 pixel-grid loader replaces the wave loader on the chat surface,
the working label shimmers, and the live status above the composer shows the
elapsed time of the current tool. Thinking: the streaming assistant row renders
the reference Thinking trace — open while the run is active with a shimmering
"Thinking" / "Searching the web" / "Running tools" header, reasoning
paragraphs, tool rows and search results fading up onto a guide line as they
arrive — and settles to a collapsed "Thought for Ns" / "Searched the web" /
"Ran N tools" header. Tool rows gain the compact label + mono detail layout
the trace relies on. Restores upstream's sandbox tool categorisation helpers in
the timeline so unified sandbox calls keep their labels and output.

## Demo

First: the pixel-grid loader, shimmer label and elapsed counter while a command runs. Second: the trace opens live with "Thinking" then "Running tools", tool rows fade in as they run, and it settles to "Thought for Ns" / "Ran 4 tools".

## Verification

Typecheck, the full `plugins/web-ui` suite, eslint and prettier pass at this level of the stack (each level is verified on its own, on top of the previous one). Live QA of the finished stack ran in a browser-only `dev-instance` against a seeded conversation (tool run, markdown table, TypeScript and diff fences, a pending approval, ⌘K, list pages, settings), in light and dark and at phone width.

## Stack

Merge in order; each PR is based on the one above it and retargets automatically when its base merges.

| # | PR | Change |
| --- | --- | --- |
| 1 | #1053 | Beautiful UI foundation — tokens, fonts, styles scaffold |
| 2 | #1054 (this) | loading state and the live thinking trace |
| 3 | #1055 | chat rows, session header and empty state |
| 4 | #1056 | approval card with a pager |
| 5 | #1057 | tool chips and code blocks |
| 6 | #1058 | records tables and filter lists |
| 7 | #1059 | search palette |
| 8 | #1060 | settings as inspector cards |
| 9 | #1061 | selection actions toolbar |
| 10 | #1062 | recommendation card |

PR #1053, opened 2026-09-10
web-ui: Beautiful UI foundation — tokens, fonts, styles scaffold

time-attack:
## Summary

Defines the Beautiful UI palette (page / canvas / surface / inset / ink / line /
field, blue / green / orange / red with tint and ink variants, tooltip colours),
the 6/8/10/14 px radius scale, the strong ease and the shared keyframes in
shell.css, and maps the existing shadcn tokens onto it so every surface
recolours consistently in light and dark. Self-hosts Inter and JetBrains Mono
(latin subsets, OFL notice) through Vite. Creates one src/styles/<component>.css
per upcoming primitive, each imported from shell.css and scanned by the css-vars
guardrail. Imported custom themes emit the same tokens. Adds the shared jsdom
harness the component tests use, and teaches the brand guard to skip font
binaries the way it skips images.

## Demo

Light and dark toggled on a live conversation: the palette, Inter and JetBrains Mono, and the shared radius scale recolour every surface consistently.

## Verification

Typecheck, the full `plugins/web-ui` suite, eslint and prettier pass at this level of the stack (each level is verified on its own, on top of the previous one). Live QA of the finished stack ran in a browser-only `dev-instance` against a seeded conversation (tool run, markdown table, TypeScript and diff fences, a pending approval, ⌘K, list pages, settings), in light and dark and at phone width.

## Stack

Merge in order; each PR is based on the one above it and retargets automatically when its base merges.

| # | PR | Change |
| --- | --- | --- |
| 1 | #1053 (this) | Beautiful UI foundation — tokens, fonts, styles scaffold |
| 2 | #1054 | loading state and the live thinking trace |
| 3 | #1055 | chat rows, session header and empty state |
| 4 | #1056 | approval card with a pager |
| 5 | #1057 | tool chips and code blocks |
| 6 | #1058 | records tables and filter lists |
| 7 | #1059 | search palette |
| 8 | #1060 | settings as inspector cards |
| 9 | #1061 | selection actions toolbar |
| 10 | #1062 | recommendation card |

PR #1059, opened 2026-09-10
web-ui: search palette

time-attack:
## Summary

The ⌘K palette takes the reference look: a scrim that darkens in both themes,
a 560 px surface with hairline dividers, an input row with the search glyph,
result rows with an avatar mark, blue highlight marks, mono tabular times and
kbd chips, group headers, a shimmering "Searching…" state and the reference
empty copy. Keyboard behaviour is unchanged.

## Demo

⌘K, typing "churn" letter by letter, arrowing through results, Escape.

## Verification

Typecheck, the full `plugins/web-ui` suite, eslint and prettier pass at this level of the stack (each level is verified on its own, on top of the previous one). Live QA of the finished stack ran in a browser-only `dev-instance` against a seeded conversation (tool run, markdown table, TypeScript and diff fences, a pending approval, ⌘K, list pages, settings), in light and dark and at phone width.

## Stack

Merge in order; each PR is based on the one above it and retargets automatically when its base merges.

| # | PR | Change |
| --- | --- | --- |
| 1 | #1053 | Beautiful UI foundation — tokens, fonts, styles scaffold |
| 2 | #1054 | loading state and the live thinking trace |
| 3 | #1055 | chat rows, session header and empty state |
| 4 | #1056 | approval card with a pager |
| 5 | #1057 | tool chips and code blocks |
| 6 | #1058 | records tables and filter lists |
| 7 | #1059 (this) | search palette |
| 8 | #1060 | settings as inspector cards |
| 9 | #1061 | selection actions toolbar |
| 10 | #1062 | recommendation card |

### DIFF
--- a/plugins/web-ui/src/search.ts
+++ b/plugins/web-ui/src/search.ts
@@ -256,7 +256,13 @@ function resultRows(): TemplateResult[] {
           <span class="chat-search-snippet" dir="auto">${highlight(hitSnippet(hit))}</span>
           <span class="chat-search-meta"
             ><bdi>${hit.entryType === "user" ? (hit.author ?? "you") : "agent"}</bdi> ·
-            ${new Date(hit.createdAt).toLocaleDateString()}</span
+            ${
+              Number.isFinite(hit.createdAt)
+                ? html`<time datetime=${new Date(hit.createdAt).toISOString()}
+                    >${new Date(hit.createdAt).toLocaleDateString()}</time
+                  >`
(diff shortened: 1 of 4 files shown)


time-attack:
Closing per review: holding off on importing Beautiful UI wholesale for now (interactions read as too stimulating and too Conductor-like). A subset may come back as our own take in collaboration with Eve. Branch `bui/07-search` is kept.

PR #1067, opened 2026-09-10
Let agents hand off emails for review and send in the chat

time-attack:
Agents can hand a finished email to the person for review and send, right inside the chat. The agent calls `send_email`; a quiet card with To / Cc / Subject / body and a Send button appears in the conversation. Nothing is sent until the person clicks Send, and the send goes out through their own Gmail connection.

## Screenshots

The real card against the app's stylesheet with a mocked ledger item (light and dark). Files the agent attached show as plain links to the stored file.

| Light | Dark |
| --- | --- |
| light | dark |

## Changes since review

The first cut was heavyweight in both the UI and the diff. This revision keeps the flow and drops the weight, roughly halving the diff (+1865/−32 to +926/−24 across 28 files):

- **Removed:** the Preview/Edit toggle and in-card editing (to change the draft, ask the agent; it files a fresh card), attachment-chip remove controls, Discard → Reopen, the collapse-to-receipt animation, the delivery-stream refresh wiring, and the markdown-to-HTML email renderer in `plugins/chassis`. Bodies go out as the plain text the agent wrote, as inbox reply drafts always have.
- **Card:** a self-contained custom element (`<email-draft-card>`) that fetches its own ledger item, so the chat surface gained five lines and no redraw hooks or cache-key changes. About 50 lines of CSS instead of 256.
- **Server:** unchanged in shape, minus the HTML part of the MIME. Sends carry no proposal or revision, because without human edits the held draft is the only version that exists.

## Example

**What the person types**

> Email dana@northwind.io about the Q3 pricing question, cc Priya, keep it short, and attach the Q3 sheet.

**What the agent calls**

```json
{
  "name": "send_email",
  "input": {
    "to": ["dana@northwind.io"],
    "cc": ["priya@acme.co"],
    "subject": "Q3 pricing for Northwind",
    "body": "Hi Dana,\n\nShort answer: no change for existing seats. The 12% uplift applies only to seats added after October 1; the sheet is attached.\n\nSina",
    "attachments": ["q3-pricing.csv"]
  }
}
```

**What the tool returns to the agent**

> Draft `<itemId>` is held for the user's review. The conversation now shows it with Send and Discard buttons, and only the user can send it; a send from an agent is refused.

**What the person sees:** the card above. Send posts `{"kind":"send"}` to the ledger item; the card turns into a one-line "Sent to dana@northwind.io" receipt. Discard turns it into "Email draft discarded".

| Situation | What happens |
| --- | --- |
| Gmail not connected for this user | The tool is not offered; if called anyway it explains how to connect under Keychain |
| Recipient, subject, or body missing or malformed | The tool refuses before anything is filed |
| Attachment file missing, or more than 10 | The tool refuses with the path or the limit |
| Attachments over 5 MB total at send time | Send fails with "attachments exceed 5 MB in total" |
| Agent tries to send via the ledger API | HTTP 403 `human_required` |
| Draft pruned from the ledger (old chat) | The card reads "This email draft is no longer available." |

## How it works

- **`send_email` tool** (`src/harness/agent-tools.ts`). Validates recipients, subject, body and attachment paths; stages attachments through the same outbound collector the `attach` tool uses, on a fresh seed per call; files the draft as a held `compose` item in the person's Inbox ledger via `holdEmailDraft` (`src/loops/email-draft.ts`); returns `display.emailDraft = { loopId, itemId }` for the card. Offered only on web conversations where the user's Gmail token resolves.
- **Ledger, not HiLO.** The ledger already has send, dismiss and per-item ownership, so the card reuses it. The action route refuses a send on a compose item from any agent capability; only the person sends. Compose items are excluded from the Inbox sync playbook (task version bump) and from the Inbox pane.
- **Gmail adapter** (`src/loops/sources/gmail.ts`). Plain-text body as before; with attachments the message is multipart/mixed. Attachments are opened as the loop owner through viewer-scoped file access, and the stored artifact's own name and type are what go out. Compose items send without threading headers and without a "Re:" prefix.
- **Card** (`plugins/web-ui/src/email-draft-card.ts`). A custom element that GETs its ledger item on connect and POSTs `send` or `dismiss`; on a 409 it reloads the item, so a draft sent from another tab settles into its receipt.

## Deliberately not included

In-card editing, "ask agent to change" button, a split Send menu, and a Slack surface for the card. On Slack the tool is not offered and the existing Gmail-drafts skill path applies.

## Testing

- Core: `test/email-draft.test.ts` (holder, compose subject and plain-text MIME, multipart with attachments, artifact-gone and oversize paths, tool validation and staging, tool gating), `test/loop-item-routes.test.ts` (agent capability refused on compose send; attachments opened as the owner).
- Web UI: `plugins/web-ui/test/email-draft.test.ts` (extractor; jsdom card renders To/Cc/Subject/body and the attachment link, sends nothing on render, Send posts once and the card becomes a receipt), `inbox-server-route.test.ts` (ledger item paths relay for every signed-in user).
- Earlier increments were verified live in a dev instance with a connected Gmail account (agent drafts with an attachment, card renders and survives the turn settling, send delivers). The slimmed card was checked against the mocked ledger in the screenshots above.


time-attack:
Three fresh-context reviews (core/harness, web UI + server gate, ledger/Gmail semantics) ran against the first commit. Resolved in the two follow-up commits:

- The Inbox sync playbook would have closed compose drafts as "already replied" because their From line is the user. The playbook now names compose items as outgoing mail under review and the task version bumps to 4 so existing crons pick it up.
- Compose items stay out of the Inbox pane rather than appearing as mail from the user to themselves. The chat card is their surface.
- The draft holder prunes the ledger like the ingest route, so drafts do not accumulate for users without Inbox sync. A draft that has aged out renders a static receipt instead of an error.
- The card serializes saves and sends per item: a save in flight no longer disables Send, text typed while a save lands is kept, and a Send clicked right after leaving a field waits for that save and reuses its revision. Conflicts rebase only inside the 409 handler and use the Inbox's wording.
- The ledger refuses a "send" on a compose item from any agent capability, so "only the person sends" is enforced at the route every path flows through. Ordinary inbox replies are unchanged.
- The tool is gated on the Gmail connector token the orchestrator actually resolved, not on the env map.
- OpenCode registers the `send_email` definition so the per-turn enable map can reach it.
- `Name <addr>` recipients pass validation. Skill guidance no longer tells agents to both show the text and call the tool.

Noted, not changed: if only a company-type Google account is connected, the send goes out from it while the card shows the signed-in address as From.

time-attack:
Follow-up increment: the agent can attach workspace files (`attachments` on `send_email`, up to 10, 5 MB total) and the body supports light markdown (bold, italic, code, links, simple lists) that goes out as multipart/alternative with a plain-text part and an HTML mirror. The card previews the body through the same converter the email uses and shows attachments as chips that can be dropped in edit mode.

A fresh review of that increment caught a real hole, fixed in the last commit: the send path opened attachment artifacts by bare id, so an edited draft could have emailed another user's file under an innocent name. Attachments are now opened as the loop owner through the viewer-scoped file access, and the artifact's own name and type are what go out. Also fixed: artifact id collisions between the chat's attach tool and the email staging (each email call now stages on its own seed), MIME header injection via a forged mimetype, RFC 2231 filenames, the size cap being checked after buffering, and a preview that rendered more than the email would.

time-attack:
Recording the live demo caught one more real bug, fixed in the latest commit: the inbound security screen sometimes quarantined `send_email`'s own result text ("Tell them it is ready and stop" reads like an injected instruction), which parked the hand-off behind an approval card. The result is core-authored and is now marked as such, with factual wording. The live GIF in the description is the run after that fix: draft with markdown and an attachment, card, send, receipt, against a real Gmail connection.

time-attack:
Slimmed it down in response to the heavyweight feedback. The diff is now +926/−24 across 28 files (was +1865/−32 across 34), and the card is a quiet inline block: To / Cc / Subject, the body as written, attachments as plain file links, Send and Discard.

Cut: the Preview/Edit toggle and in-card editing (ask the agent to change it; it files a fresh card), attachment-chip remove controls, Discard → Reopen, the collapse-to-receipt animation, the delivery-stream refresh wiring, and the markdown-to-HTML email renderer in `plugins/chassis` (bodies go out as plain text, as inbox reply drafts always have). The card is now a self-contained custom element that fetches its own ledger item, so `chat.ts` gained five lines and no redraw hooks; CSS went from 256 lines to about 50. Sends carry no proposal or revision, since without human edits the held draft is the only version that exists. Tests were rewritten for what remains.

Fresh screenshots (light and dark) are in the PR body.

PR #1042, opened 2026-09-10
Scope ambient policy to supported surfaces

time-attack:
## Summary

- keep web-project standing orders available for addressed turns while removing unsupported ambient/bot controls from member and admin editors
- derive policy applicability on the server and enforce it consistently across member, admin, and agent-facing writes
- reject unsupported web-project `ambientEnabled`/`bots` fields atomically without deleting legacy values or history
- make web-project tool guidance and the composed autonomous-mode prompt explicitly addressed-only while preserving Slack autonomous behavior

Fixes #317.

## Test plan

- prompt/resolution/orchestrator suites: 190 passed
- policy/project/auth/core suites: 382 passed
- web UI suites: 32 passed
- admin plugin suites: 114 passed
- root, web UI, admin, CLI, and contract typechecks
- targeted ESLint, Oxlint, Prettier, and `git diff --check`

The tests cover orders-only save/read and prompt use, atomic rejection, legacy preservation, stale UI responses, full composed prompt semantics, and Slack channel/group-DM controls.

## Before / after

Light theme, mock data. Web project scope is `group:web-project-launch-plan`; the Slack scope is a channel.

### Web UI, project settings (web project)

| Before | After |
| --- | --- |
| Web UI ambient policy editor for a web project before: ambient toggle, standing orders, automated posters | Web UI standing orders editor for a web project after: only standing orders remain |
| Web projects offered an ambient toggle and a bot ledger that nothing on this surface honored. | Only standing orders remain, framed as instructions for addressed replies. |

### Admin, scope governance (web project)

| Before | After |
| --- | --- |
| Admin ambient reply policy card for a web project before: channel copy, ambient behavior select, bot ledger | Admin standing orders card for a web project after: heading, hint, and standing order textarea only |
| Channel-oriented copy, ambient behavior select, bot ledger, and judgment log link for a web project. | Card retitled to standing orders with a project hint; unsupported controls and the judgment log link are gone. |

### Slack channel scope (unchanged on this branch)

| Web UI | Admin |
| --- | --- |
| Web UI agent behavior editor for a Slack channel after: ambient toggle, standing orders, and bot ledger still present | Admin ambient reply policy card for a Slack channel after: full controls still present |
| Ambient toggle, standing orders, and automated posters still editable. | Ambient behavior, standing order, bot ledger, and judgment log link still present. |

---
[...]

Issue #825, opened 2026-08-28
Triage: fix(admin): correct ambient-policy help text to match actual default rule


time-attack:
Tracking the change proposed in #527 (by @SomSamantray).

We're moving this from a code PR to a tracked issue under [`CONTRIBUTING.md`](https://github.com/yc-software/qm/blob/main/CONTRIBUTING.md): feature ideas belong in `adrs/` as a short human-written note and bugs as an issue, and we implement the underlying code from our side. Capturing the original proposal here so it isn't lost.

**Original PR:** #527
**Reported by:** @SomSamantray

---
> ## Summary
> 
> The admin panel's "Ambient behavior" help text claimed ambient mode defaults on/off based on channel member count ("on for channels with 8 or fewer members, off for larger ones"). No such logic exists anywhere in the codebase — it's a fictional rule. This PR corrects the copy (in two places) to describe the actual default behavior.
> 
> ## Root cause
> 
> The real default logic lives in `judgeAmbientContainer` (`src/api/app-ambient.ts:89-96`): when a channel's `ambientEnabled` policy is unset ("Default"), ambient is **off** unless the channel has a non-empty standing order (`policy.orders`) or at least one bot configured with `mode: "action"`. There is no member-count check anywhere — confirmed by reading the actual logic and grepping the repo for any member-count-based ambient rule.
> 
> ## Changes
> 
> - `plugins/admin/public/index.html`:
>   - Per-channel "Ambient behavior" hint text and the `Default (by channel size)` dropdown option — corrected to describe the standing-order / action-bot condition.
>   - Org-wide "Allow ambient behavior" toggle hint — caught during review, since it repeated the same fictional "8 members" claim in a second location.
> 
> Copy-only change; `src/api/app-ambient.ts` and all other logic are untouched.
> 
> ## Validation
> 
> - Re-read `judgeAmbientContainer`'s `ambientOffReason` branches line-by-line against the new copy to confirm every claim is directly traceable to the code.
> - Grepped the repo for any remaining member-count-based ambient text/logic — none found.
> - Ran the changed HTML through Python's `html.parser` to confirm it still parses cleanly.
> - No dev-server/browser render was performed (this sandbox has no reachable browser driver and the full dev instance requires Postgres/service setup not available here); the change is plain text content with no new interactive elements.
> 
> Fixes #270
> 
> 
> ---
[...]

PR #1081, opened 2026-09-10
Correct the Admin ambient-policy default guidance

### LINKED ISSUE #270: Admin's ambient help text describes a channel-size rule the code doesn't implement

time-attack:
Fixes #270.

## Summary
- Correct the three admin-UI copy strings claiming the ambient reply default is decided by channel size (an "8 or fewer members" / "more than 8 members" cutoff that doesn't exist in the runtime). The actual default (`src/api/app-ambient.ts`, `judgeAmbientContainer`) opts an unset per-channel policy into ambient only when there's a non-empty standing order or a bot ledger entry with mode `"action"`; an org-wide off always wins, including over an explicit per-channel On.
- Fix the matching stale assertion message in the channel-policy-store tri-state test (the asserted value was already correct; only the description text repeated the size-rule framing).
- Add an admin copy-regression test asserting the corrected strings and the absence of any remaining "8 members" / "channel size" language.

No runtime/behavior change. Reported by @ianTPE in #270 (a 5-member Slack channel that the Admin UI promised automatic ambient behavior for, but that never engaged — the real gate is standing orders/action bots, not member count). Same underlying stale-copy issue as the earlier, unmerged #825.

## Coordination with #1042
#1042 restructures the same `card-ambient-policy` HTML block (for web-project applicability) but its diff leaves this stale copy text itself unchanged, and it doesn't touch the org-wide card or either test file here. Not modifying #1042. Recommend landing #1042 first and rebasing this on top of it — the two per-channel-card string edits move into #1042's new wrapper markup, and the org-card edit plus both test files carry over with no conflict either order.

## Test plan
- `node --test test/surface-cache.test.ts` — 21/21 pass
- `node --test test/surface-cache-ambient.test.ts` (full ambient policy matrix: org-off precedence, per-channel off, standing-orders-opts-in regardless of size, action-bot opts-in, explicit-on opts-in, bot ledger semantics) — 40/40 pass
- `plugins/admin`: `node --test` (full plugin suite incl. new copy test) — 130/130 pass
- Root typecheck, admin plugin typecheck, eslint, oxlint, prettier --check, git diff --check — all clean
- Confirmed the new copy-regression test actually catches the regression: fails against the pre-fix HTML, passes against the fix
- Independently reviewed against a fresh-context pass per repo convention
- Demoed against the actual Admin UI running locally against a mock harness with in-memory data (before/after screenshots to follow in review)

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/1081"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791670164&installation_model_id=19911&pr_number=1081&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F1081&signature=197399c99aec0bd7dc74df189f3eb05b402694d8af8b2b07d448e1d3626605a1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/admin/public/index.html
+++ b/plugins/admin/public/index.html
@@ -5418,10 +5418,11 @@
                     <label for="ambient-enabled">Ambient behavior</label>
                     <p class="hint" style="margin: 2px 0 6px">
                       When off, the agent never wakes on overheard messages in this channel; it only responds to direct
-                      @mentions. Default: on for channels with 8 or fewer members, off for larger ones.
+                      @mentions. Default: on when this channel has a standing order or an action-marked bot, off
+                      otherwise.
                     </p>
                     <select id="ambient-enabled" style="margin-bottom: 14px">
-                      <option value="default">Default (by channel size)</option>
+                      <option value="default">Default (standing order or action bot)</option>
                       <option value="on">On</option>
                       <option value="off">Off</option>
                     </select>
@@ -5553,8 +5554,8 @@
                       <span class="setting-copy"
                         ><strong>Allow ambient behavior</strong
                         ><small
-                          >Channels choose per-channel; without a choice, ambient defaults off in channels with more
-                          than 8 members.</small
+                          >Channels choose per-channel; without a choice, ambient defaults on only where there's a
+                          standing order or an action-bot trigger, off otherwise.</small
                         ></span
                       >
                     </label>
--- a/plugins/admin/test/ambient-policy-copy.test.ts
+++ b/plugins/admin/test/ambient-policy-copy.test.ts
@@ -1,0 +1,28 @@
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import test from "node:test";
+
+const html = readFileSync(new URL("../public/index.html", import.meta.url), "utf8");
+
+test("ambient policy copy: per-channel default explains the real gate — standing orders or an action bot, not channel size", () => {
+  assert.match(html, /Default: on when this channel has a standing order or an action-marked bot,\s+off\s+otherwise\./);
+  assert.match(html, /<option value="default">Default \(standing order or action bot\)<\/option>/);
+});
+
+test("ambient policy copy: org-wide toggle explains the same default and that it overrides an explicit per-channel On", () => {
+  assert.match(
+    html,
+    /Org-wide, default on\. When off, the agent never acts on overheard messages in any channel,\s+regardless of per-channel settings; it only responds to direct @mentions\./,
+  );
+  assert.match(
+    html,
+    /ambient defaults on only where there's a\s+standing order or an action-bot trigger, off otherwise\./,
+  );
+});
+
+test("ambient policy copy: no stale channel-size language remains anywhere in the admin UI", () => {
+  assert.doesNotMatch(html, /8 or fewer members/);
+  assert.doesNotMatch(html, /by channel size/);
+  assert.doesNotMatch(html, /more\s+than 8 members/);
+  assert.doesNotMatch(html, /channels with more/);
+});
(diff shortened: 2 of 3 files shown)


time-attack:
### Rendered before and after

Actual Admin UI running locally with a mock harness and in-memory test data. No production data or real Slack workspace was used.

**Before:** the control promises an automatic channel-size default.

Before: misleading channel-size default

**After:** the control names the real standing-order/action-bot default. Runtime behavior is unchanged.

After: correct standing-order/action-bot default

PR #1085, opened 2026-09-10
Link quarantine refusals to authorized review history

### LINKED ISSUE #574: Quarantine refusal points at a review flow that isn't reachable; tightening posture to `strict` silently removes all content screening
# Quarantine refusal points at a review flow that isn't reachable; tightening posture to `strict` silently removes all content screening

Two small gaps found while exercising `guidance` standing-order changes over
Slack on a self-hosted deployment. Neither is about the screen's judgement — a
model-based classifier having a false-positive rate is a trade-off, not a bug.
Both are about what an operator can do once it fires.

Context: across 61 standing-order updates in one channel, all posted by an org
admin with a direct @mention, 12 (19.7%) were refused by the screen. So this path
gets exercised in ordinary use, not just adversarially.

## 1. The refusal names a review process that isn't wired up

```
I couldn't act because my security screen flagged part of this message or its
conversation context. Please retry without the flagged context, or ask an admin
to review the quarantine.
```

`SECURITY_QUARANTINE_REFUSAL_TEXT` in `plugins/chassis/src/security-quarantine.ts`
is only ever rendered by `src/delivery/run-result-delivery.ts` and
`src/slack/refusals.ts`; nothing else consumes it outside its test.

The flagged material does survive, though — what's absent is any link an admin
could follow:

- On quarantine the orchestrator appends the refused input into the session
  transcript with `securityTainted: true` (imported overheard messages too —
  `src/core/orchestrator.ts` ~692-727), and the screen's full request snapshot
  is recorded against the same session.
- An admin can read both back: `GET /v1/admin/sessions/:id/llm?turnSeq=N`
  returns the screen request (`src/api/routes/admin/sessions.ts` ~274; the
  parameterless variant omits it).

But the `security_posture.quarantine` audit row records neither the sessionId
nor anything that locates the payload, and unlike other refusal kinds the
quarantine path returns before an `adminUrl` can be attached —
`src/slack/refusals.ts:33-35` substitutes the fixed string with no reference.
So the review the sentence promises is reachable with the storage that already
exists; only the wire-up is missing. Either the sentence should stop promising
a review, or the refusal/audit row should link to the session holding the
payload.

## 2. `strict` is described as tightening but turns content screening off

The admin resource labels posture as *"The org value is a minimum; narrower
scopes may tighten it but cannot weaken it."* But:

```ts
// src/security/security-posture.ts
dangerous: { inboundScreening: "off",      toolApprovals: "none" }
auto:      { inboundScreening: "external", toolApprovals: "none" }
strict:    { inboundScreening: "off",      toolApprovals: "all"  }
```

`auto` is the only posture that screens inbound content. Moving a scope from
`auto` to `strict` — the documented way to tighten — **removes** the inbound
screen and substitutes an approval prompt on every tool call. And it isn't only
the user-input screen; `strict` disables every screening path:

- user-input screening (`src/core/orchestrator.ts` ~611)
- steer screening, which returns `"allow"` before screening (`~312`)
- tool-result screening (`~2089`)
- attachment screening (`~554`)

Verified on this deployment: a channel set to `dangerous` resolves back to
`auto` as the floor rule says, while `strict` sticks and the quarantine path
stops firing.

The trade may well be deliberate — the strict text in `renderSecurityPolicyPrompt`
reads as approval-gating replacing screening — but the label's "cannot weaken
it" reads as a per-control guarantee, and the postures don't form a lattice:
`auto`'s `inboundScreening: "external"` is not below `strict`'s `"off"`. An
operator tightening a scope in response to a security concern would not expect
to lose a control by doing so. At minimum the label seems worth qualifying.

## Possibly related

`src/slack/conversation.ts:195` builds `overheard` from every displayed message
regardless of author, so an admin's own earlier addressed request can reach the
screen as overheard data. Two qualifications after reading closer:

- Replay is deduplicated: `selectOverheardToImport` drops messages already
  imported into the session, so prior requests that successfully triggered a
  turn don't come back. It's session creation and never-imported messages
  that can.
- The provenance gap is narrower than "present but unused":
  `renderOverheard` embeds `from="human"` and the author inside the rendered
  content, but the screen payload's source label is the bare `"overheard"`
  (`src/core/orchestrator.ts:605`), and the screen prompt grants direct
  human-context status only to sources named `sender` or ending in
  `:unprompted`. `securityScreenPayload` even has an overheard branch that
  would label entries `overheard:<name>`
  (`src/security/security-posture.ts:131`) — but the user-input call site
  passes `overheard: []` (`orchestrator.ts:615`) and routes overheard through
  `externalPromptData` instead, so that branch never runs on this path.

That would explain why quarantine reasons cite channel history rather than the
request itself. Offered as an observation only: I have no evidence it accounts
for the rate, and the verdicts are not reproducible from request density
either — the densest stretch (12 requests in 7 minutes) was entirely clean,
while isolated requests after long gaps were sometimes refused.

Happy to share the harness that produced this.

time-attack:
## Summary

- link quarantine refusals to authorized session history across direct Slack, web, approval-resume, and addressed async-recovery paths
- canonicalize and validate HTTP(S) review URLs, resolve legacy recovery thread references, and render safe web links without exposing internal screening reasons
- accurately describe Auto, Strict, Dangerous, and command-policy behavior

Security enforcement is unchanged; this only corrects user-facing copy, review handoff wiring, safe rendering, and posture documentation.

Fixes #574, reported by @ianTPE.

## Verification

- 104 affected root tests passed
- 30 affected web tests passed
- root, contract, and web typechecks passed
- ESLint, oxlint, Prettier, and diff checks passed
- independently reviewed and approved after two review passes

Real Slack QA was not performed because the full dev instance was blocked by unavailable local Docker/Sprites infrastructure. A synthetic browser demo using the branch's actual shared formatters, resource registry, and safe Lit link renderer was completed instead; no production data or credentials were used.

---
<a href="https://app.blacksmith.sh/yc-software/codesmith/qm/pr/1085"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791675811&installation_model_id=19911&pr_number=1085&repository=yc-software%2Fqm&return_to=https%3A%2F%2Fgithub.com%2Fyc-software%2Fqm%2Fpull%2F1085&signature=32e78f9b5f19bc821e02a5355c594e7dffbac5ea9453ac62e2c16fc47124e1aa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/plugins/chassis/src/failure-copy.ts
+++ b/plugins/chassis/src/failure-copy.ts
@@ -1,4 +1,4 @@
-import { SECURITY_QUARANTINE_REFUSAL_TEXT } from "./security-quarantine.ts";
+import { quarantineRefusalText } from "./security-quarantine.ts";
 
 export const GENERIC_FAILURE_CLAUSE = "something went wrong on my end";
 
@@ -8,10 +8,11 @@ export interface FailureLike {
   status?: string;
   reason?: string;
   refusalKind?: string;
+  adminUrl?: string;
 }
 
 export function userFacingFailureText(result: FailureLike): string {
-  if (result.refusalKind === "security_quarantine") return SECURITY_QUARANTINE_REFUSAL_TEXT;
+  if (result.refusalKind === "security_quarantine") return quarantineRefusalText(result.adminUrl);
   if (result.status === "refused" && result.reason) return result.reason;
   return GENERIC_FAILURE_TEXT;
 }
--- a/plugins/chassis/src/security-quarantine.ts
+++ b/plugins/chassis/src/security-quarantine.ts
@@ -1,2 +1,19 @@
-export const SECURITY_QUARANTINE_REFUSAL_TEXT =
-  "I couldn't act because my security screen flagged part of this message or its conversation context. Please retry without the flagged context, or ask an admin to review the quarantine.";
+const SECURITY_QUARANTINE_REFUSAL_INTRO =
+  "I couldn't act because my security screen flagged part of this message or its conversation context.";
+
+export const SECURITY_QUARANTINE_REFUSAL_TEXT = `${SECURITY_QUARANTINE_REFUSAL_INTRO} Please retry without the flagged context.`;
+
+export function quarantineRefusalText(adminUrl?: string): string {
+  if (!adminUrl || /[\u0000-\u001f\u007f]/.test(adminUrl)) return SECURITY_QUARANTINE_REFUSAL_TEXT;
+  const reviewUrl = adminUrl.trim();
+  if (!reviewUrl || /[\s<>"']/.test(reviewUrl) || !/^https?:\/\//i.test(reviewUrl))
+    return SECURITY_QUARANTINE_REFUSAL_TEXT;
+  try {
+    const url = new URL(reviewUrl);
+    if ((url.protocol !== "http:" && url.protocol !== "https:") || !url.hostname)
+      return SECURITY_QUARANTINE_REFUSAL_TEXT;
+    return `${SECURITY_QUARANTINE_REFUSAL_INTRO} Please retry without the flagged context, or ask an admin to review this quarantine in session history: ${url.href}`;
+  } catch {
+    return SECURITY_QUARANTINE_REFUSAL_TEXT;
+  }
+}
--- a/plugins/web-ui/src/chat.ts
+++ b/plugins/web-ui/src/chat.ts
@@ -95,7 +95,7 @@ import { deepLinkPath, UI_BASE } from "./deep-link";
 import type { ChatSurface, ConvCtx } from "./conv-types";
 import { errMessage, swallow } from "../../chassis/src/errors";
 import { showStateError } from "./error-banner";
-import { splitLinks } from "./linkify";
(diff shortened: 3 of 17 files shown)


time-attack:
### Rendered quarantine-review handoff

This uses the branch's actual shared formatters and safe web link renderer with synthetic local data. It is a rendering/wiring demonstration, not a screenshot of a real Slack workspace. Direct refusal, addressed recovery, stored web refusal, and the no-admin-URL fallback are shown.
quarantine-review-handoff

PR #1125, opened 2026-09-11
Rename projects from the Projects page

### LINKED ISSUE #321: No way to rename a project from the Projects page
The Projects page is where you manage a project — you can add and remove members there —
but there's no way to rename one. I went looking for it there, found nothing, and concluded
that projects couldn't be renamed at all.

The only rename entry I could find is on the Chats page: the `⋯` menu on a project row in
the sidebar has a Rename item (owner only). That works fine, it's just not where you'd think
to look.

Would it make sense to expose rename on the Projects page too, alongside the member
management?

time-attack:
## Summary

Adds an owner-only inline rename affordance to the project title on the Projects detail page.

Fixes #321, reported by @ianTPE.

- Reuses the existing `renameProject` action (existing `PATCH /v1/projects/:id`, existing owner-checked authorization) — no new backend, no new capability.
- Small inline editor next to the project title, subtle pencil affordance for the owner only, matching the existing Slack-channel-linker editor pattern.
- Keyboard: Enter saves, Escape cancels. Blank names show a validation error and stay in edit mode; unchanged names exit quietly. A busy guard disables the controls during a save, and a failed save preserves the typed draft with an error instead of discarding it silently.
- Navigation-race safe: a slow save can't clobber a different project's draft/error after navigating away, or a freshly reopened draft on the same project.
- Focus is restored predictably after every exit path (including on a real async save/failure), and never stolen from elsewhere on the page.
- Visible on touch as well as desktop hover/focus.

## Verification

- 1026 web-ui tests pass
- Typecheck, ESLint, oxlint, and Prettier all clean
- Independently reviewed before this PR was opened

## Demo

Rendered locally against synthetic fixture data (no production data), verified via automated real-DOM tests plus manual screenshots taken outside this repo.

### DIFF
--- a/plugins/web-ui/src/contexts.ts
+++ b/plugins/web-ui/src/contexts.ts
@@ -105,6 +105,12 @@ export const contextsState = {
   titleValue: "",
   titleBusy: false,
   titleError: "",
+  // Bumped whenever the title-edit session is torn down or a new one begins (reset,
+  // navigating to a different scope, opening a fresh edit). A commit in flight captures
+  // this generation plus the project id it's editing, so a late response from an old
+  // session can never overwrite a newer or different one (including A -> B -> A again).
+  titleEditGen: 0,
+  titleEditProjectId: null as string | null,
 };
 
 let contextsLoading = false;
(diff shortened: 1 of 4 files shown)


time-attack:
### Actual UI preview

Rendered from this branch’s real QM UI with synthetic local data, cropped to the project header. The images show the owner’s rename affordance, inline editing, and the saved result. No production data was used.
rename-1-pencil
rename-2-editing
rename-3-saved

Issue #951, opened 2026-09-04
Custom-provider models always get reasoning_effort + tools → OpenAI rejects gpt-6-astra on /v1/chat/completions


trackport-dev:
**Version:** CLI 0.1.7 (core image from the 0.1.7 manifest), pi-ai 0.82.0. Target: Fly. Harness: `pi`.

## What happened

I registered a custom provider pointing at OpenAI itself so I could use a model that isn't in the built-in registry yet:

```
PUT /v1/admin/custom-providers/openai-direct
{ "name": "OpenAI Direct", "protocol": "openai", "baseUrl": "https://api.openai.com/v1",
  "models": [{ "id": "gpt-6-astra", "name": "GPT-6 Astra", "contextWindow": 1000000, "maxTokens": 128000 }],
  "apiKey": "sk-…" }
```

Registration and key validation succeed and the model appears in the picker. Every turn on it then fails:

```
400: {"message":"Function tools with reasoning_effort are not supported for gpt-6-astra in /v1/chat/completions.
To use function tools, use /v1/responses or set reasoning_effort to 'none'.","type":"invalid_request_error","param":"reasoning_effort"}
```

The same key + model works from a plain `POST /v1/chat/completions` (with and without `stream`) when no `reasoning_effort` is sent, and from `/v1/responses`, so the key and model access are fine. The request QM builds is what OpenAI rejects: it always includes tools, and it includes a `reasoning_effort`.

## Why (as far as I can tell)

- `toRuntimeModel()` sets `reasoning: false` (src/model/custom-providers.ts:107), but `customModelsJson()` (src/model/custom-providers.ts:161-176), which is what the pi runtime actually loads, emits only `id/name/contextWindow/maxTokens/cost` — no `reasoning`, no `compat`.
- pi-coding-agent's default thinking level is `medium`, and the pi harness only skips `applyTurnEffort` for `"auto"`, so a non-`off` effort ends up on the session for every custom model.
- Nothing an operator can reach avoids it: the web-UI effort levels (`THINKING_LEVELS` in src/model/pi-models.ts) have no `off`, and the custom-provider form has no reasoning/effort control.

Related: #847 asks for the opposite control (turn reasoning *on* for thinking models). Both point at the same gap: the custom-provider spec needs an explicit reasoning flag that is carried into `customModelsJson()` (and/or `compat.supportsReasoningEffort`), and non-reasoning custom models should get thinking `off` instead of the runtime default.

## Repro

[...]

PR #1178, opened 2026-09-14
Add explicit Responses protocol for custom providers

time-attack:
## Summary

- Add an explicit `openai-responses` opt-in for custom providers.
- Preserve the existing `openai` to Chat Completions and `anthropic` to Messages mappings and defaults.
- Carry the selected protocol through runtime models, generated Pi `models.json`, OpenCode, validation, Admin labels, and edit round trips.
- Use OpenCode's pinned `@ai-sdk/openai` adapter for Responses while legacy OpenAI custom providers remain on `@ai-sdk/openai-compatible`.

## Scope

This addresses the explicit Responses protocol follow-up posted on September 12 in #951. It is separate from the original reasoning-effort and tools report and does not change reasoning controls, tool behavior, pricing flags, existing provider IDs, or model defaults.

Protocol selection is explicit. There is no URL inference, automatic migration, or default change.

## Verification

- 75 affected tests pass on current `main`.
- A real loopback HTTP endpoint validates the key through `/v1/models` with bearer auth, then the generated Pi registry and harness issue a Responses body to `/v1/responses` and consume a streamed Responses SSE reply.
- Existing loopback tests continue to prove legacy OpenAI uses `/chat/completions` and Anthropic uses `/v1/messages`.
- The pinned OpenCode 1.17.18 adapter path was verified against its bundled `@ai-sdk/openai` 3.0.53 semantics, and the generated config regression confirms the explicit adapter selection.
- Typecheck, contract typecheck, knip, oxlint, ESLint, formatting, and diff checks pass.

## Status

Independent code review is complete and all 20 CI checks have passed. Ready for maintainer review.


time-attack:
### Admin protocol selection

Rendered from the actual Admin UI with synthetic local data. Editing the saved provider preserves OpenAI Responses as the selected protocol. No production data or real credentials were used. The separate loopback regression verifies the generated Pi runtime sends a Responses request and consumes its SSE reply.
responses-protocol-proof

PR #1222, opened 2026-09-15
web-ui: refetch a settled transcript on foreground/SSE reconnect (isolated from #1104)

time-attack:
## Fix

When a turn finishes while the browser misses its live events, foreground/SSE recovery now refetches the persisted transcript even when there is no active run left to attach. This is a read-back, not a message resend or model retry.

Only `resumeIfIdle` enables the new settled-run fallback. Initial mount behavior is unchanged. The existing refresh-generation checks discard stale results after navigation or a newer refresh; they do not deduplicate the underlying HTTP requests.

## Scope and credit

This isolates @ReganBell's terminal-reconnect hunk from #1104, with his original identity credited in the commit. It does not include that PR's mid-turn post rendering or delivery fan-out work. It reuses #1115's existing transcript-refresh safeguards.

## Verification

- Independent review approved the change.
- 30 affected tests passed: reconnect 7, confirmed-post recovery 11, steer recovery 5, inbox resync 7.
- Fail-before confirmed: four recovery scenarios fail on the base code; the two unchanged-behavior controls pass on both versions.
- App, server, and test typechecks, ESLint, and Prettier passed.
- [Actual compiled-UI visual replay](https://github.com/yc-software/qm/pull/1222#issuecomment-5687054087): a synthetic stored reply appears after the foreground signal, with no `/api/turn` request.

All 20 CI checks passed. No merge performed.


time-attack:
## Visual regression check

Actual production UI built from `2bd7e8d`, using synthetic API data. The fixture persisted the reply without delivering a live event, leaving the visible transcript stale. Replaying the foreground `visibilitychange` signal then fetched and displayed the stored reply. No `/api/turn` request or message resend occurred. This is a controlled missed-event replay, not a live phone/provider test.

### Before the foreground signal
Stored reply not yet displayed

### After the foreground signal
Persisted reply recovered

PR #1230, opened 2026-09-15
Fix custom-provider readiness across Portal and Admin

time-attack:
## Fix

Use one local serviceability check for custom-provider readiness in the Portal and Admin. An enabled provider must have a decryptable key and a model supported by the selected harness. This covers Pi and OpenCode without falsely enabling native Claude/Codex, malformed providers, or keyless configurations.

Admin onboarding now identifies custom providers and opens their existing editor instead of calling built-in-only key endpoints. Built-in Save/Disable behavior and Remove Key semantics are unchanged. No remote provider probe or secret logging was added.

## Scope and credit

The original Portal contribution from @Ntrakiyski's #1016 is retained as a separate commit with its original author. The following commit adds the shared serviceability check, Admin controls, and regression coverage. This does not duplicate #1117's discovery/classification work or alter either contributor's branch.

## Verification

- Independent functional review reproduced and cleared the Portal, OpenCode, malformed-key, and Admin-action regressions.
- 41 focused tests passed, including real HTTP and actual Admin script/DOM paths.
- Core, Admin, and Portal typechecks, lint, formatting, and Knip passed.
- [Before/after Admin visual proof](https://github.com/yc-software/qm/pull/1230#issuecomment-5687200194): the same synthetic provider changes from **Needs a key** to **Ready**, with a working Manage action.
- All 20 CI checks passed.

The final authorship-only history adjustment preserves the exact reviewed and photographed source tree (`187ce1be`). No merge performed.


time-attack:
## Visual regression check

Actual Admin HTML from `1cdad4c` and `d865d789`, served with the same synthetic keyed custom provider and the corresponding API response shapes. No real provider was contacted.

Before, onboarding incorrectly shows **Needs a key**. After, it shows **Ready**, selects the custom model, disables the built-in key field, and offers **Manage this custom provider**. I also clicked Manage and verified that the existing Acme Gateway editor opens; no key was revealed or changed.

### Before
Incorrect custom-provider readiness

### After
Correct custom-provider readiness and controls

PR #1287, opened 2026-09-16
Add persistent subagents with durable coordination and inline badges

ReganBell:
## Summary

Persistent subagents now coordinate through durable internal messages and return results to their parent without starting extra parent turns. `send_message` supports parent, child, and sibling communication; `followup_task` explicitly starts additional child work; `wait` receives messages. Tool guidance asks for one consolidated answer rather than routine completion acknowledgements.

Subagents appear as compact, clickable inline badges inside collapsed tool groups. Attached children stay out of session lists; dragging a badge into the sidebar promotes it to a standalone session. The `persistent_subagents` feature flag defaults off and supports per-person rollout from the admin editor.

Authorization is rechecked when messages are delivered, including source-entry visibility after participant membership changes. Durable deduplication, queue limits, cancellation, and restart recovery cover the session tree. Concurrent Codex tool requests no longer block child startup responses while a parent waits.

## Validation

- Live web QA with real Codex turns and Postgres: sibling message, child progress, both automatic parent results, one consolidated final response; explicit follow-up verified again after rebase. All four parent messages consumed with no automatic extra parent runs.
- Live web UI checks: inline badge navigation, sidebar promotion, hidden attached children, admin feature flag controls.
- After integrating current main: 204 affected core tests, 54 session UI tests, 11 header/cache regressions, and 65 Postgres store tests passed. Worker availability tests passed on rerun after a local timing failure.
- Root typecheck and lint passed; web typechecks and production build passed.
- Independent security and lifecycle reviews completed; findings fixed and deltas reviewed.

The feature remains disabled by default. Production rollout is separate from this merge.

## UI preview

Live browser capture using synthetic tasks and real model turns (no mocked data):

Inline subagent badges and consolidated result


ReganBell:
Live web QA after the change (real Codex turns and Postgres, synthetic tasks): compact inline child badges and one consolidated parent result. Attached children are absent from the sidebar.

clipboard

PR #1300, opened 2026-09-16
Overlap AWS web and portal updates when routing is unchanged

ReganBell:
## Change

Ordinary AWS releases currently wait for the old web-ui rollout to retire before starting portal, serializing two image updates. That ordering was introduced for the migration from a separate admin service to web-ui’s `/admin` endpoint.

Start both updates together only when the recorded task definitions and replica counts match exact stable live cohorts, portal’s web/admin upstreams are unchanged, and web’s explicit admin capability is unchanged. Missing or uncertain evidence keeps the ordered rollout. Migration ordering, final readiness checks, native deployment completion, drain durations, and compensation remain intact. Reuse the existing capacity cohort validator for this proof.

## Validation

- Five new AWS regressions: concurrent image replacement, ordered admin migration, missing settings, retiring tasks, and restoration of both workloads after web fails following portal submission.
- Three existing dependency-ordering tests and two capacity/controlled-cohort regressions pass.
- CLI typecheck, ESLint, Prettier and diff checks pass.
- Pure deployment logic tested through the AWS fixture; no runtime or UI behavior changed.
- Independent high-effort review requested before merge. No live AWS operations performed.

The change allows rollout overlap; end-to-end time saved still requires measurement.


ReganBell:
@cursor review

PR #1335, opened 2026-09-17
feat(connectors): add Microsoft Graph OAuth provider

adityathakker:
## Summary

Adds a `microsoft` entry to the OAuth `PROVIDERS` map in `src/connectors/oauth.ts`, using the existing consent/exchange/refresh/keychain pipeline every other connector (Google, Slack, Notion, Linear, Dropbox, GitHub, X) already shares — this is one more config entry, not new plumbing.

- Scopes: `offline_access`, `Mail.Read`, `Sites.Read.All`, `Files.Read` — read-only Graph delegated scopes, no write/send scope requested.
- Authorization code + PKCE (`pkce: true`), reusing the same shared token-adapter shape the GitHub and X providers already use, which rejects a 200-status OAuth error body — Microsoft's token endpoint can return `{error, error_description}` with HTTP 200 on some failure paths.
- Tenant restriction is configuration, not a hardcoded value: `authUrl`/`tokenUrl` carry a `{tenant}` placeholder resolved at request time via a new optional `tenantEnv` field on `OAuthProviderConfig`, mirroring how `GOOGLE_WORKSPACE_DOMAIN` already resolves onto `ResolvedClient.hostedDomain` for Google's company-account check. Falls back to Microsoft's multi-tenant `common` endpoint when unset, so a deployment with no configured tenant still gets a working default — and a deployment that does set it gets Microsoft's native single-tenant sign-in restriction, which is stronger than a post-hoc ID-token claim check.
- Mirrors `GOOGLE_OAUTH_CLIENT_SECRET`'s `requiredWhen` gate for `MICROSOFT_OAUTH_CLIENT_SECRET` in both the runtime secret schema (`src/deployment/secret-schema.ts`) and the standalone CLI's secret list (`cli/src/secrets.ts`), which `secret-schema-drift.test.ts` checks stay in sync.

## Test plan

- [x] `test/oauth.test.ts`: 4 new tests — tenant fallback to `common`, tenant override via `MICROSOFT_OAUTH_TENANT`, PKCE + 200-status-error-body rejection on exchange, clean refresh failure on revoked consent (no silent retry). 31/31 passing.
- [x] Full affected set (oauth, oauth-routes, oauth-consent-bridge, secret-schema-drift, connector-invariants, deploy-directory-doc): 68/68 passing.
- [x] `tsc --noEmit` clean, `eslint` clean, `prettier --check` clean.
- [x] Confirmed no other file hardcodes an OAuth provider-name enumeration that would need a "microsoft" addition — every consumer reads `PROVIDERS` dynamically.


adityathakker:
Withdrawing — deciding to keep this org's Microsoft connector implementation private for now rather than contribute it upstream.

PR #1345, opened 2026-09-17
Stream home snapshots part by part and keep the body when a rotation snapshot fails

### LINKED ISSUE #1323: AWS home snapshots fail when the sandbox lacks space for a second full archive
### Problem

The AWS sandbox snapshot path creates the complete home archive at `/tmp/agent-home.tar` before uploading it. Because `/root` and `/tmp` share the MicroVM disk, snapshotting requires enough free space for both the populated home and a second full archive. Large homes fail during `tar` with `No space left on device`, leaving changes since the previous successful snapshot unpersisted.

### Reproduction

1. Use the AWS sandbox backend with an existing durable snapshot.
2. Populate the sandbox home until remaining disk space is smaller than the resulting tar archive.
3. Trigger teardown or another home snapshot.
4. `tar -cf /tmp/agent-home.tar` fails before upload begins.

### Expected

Snapshot bytes should be streamed to multipart object storage, or otherwise persisted without materializing a second complete copy on the sandbox disk. A failed replacement must leave the previous committed object recoverable.

time-attack:
Fixes #1323

## Root cause

`createHomeSnapshotOps` (shared by the AWS, E2B, and Modal backends) built the complete home archive at the tar path (`/tmp/agent-home.tar` on AWS), measured it, then cut each part next to it with `dd` before reading it back. On the AWS MicroVM `/root` and `/tmp` share one disk, so a snapshot needed free space for the populated home plus a second full copy of it; a near-capacity home failed inside `tar` with `No space left on device` and nothing since the previous successful snapshot was persisted.

Independently, the AWS rotation path (`ensureBody`, body past `rotateAfterSeconds`) swallowed the snapshot error, reported it, and terminated the stale body anyway, discarding every change made since the last good snapshot.

## Fix

**Streaming producer (shared helper, so all three backends get it).** The snapshot script now runs `tar -cf -` in a detached pipeline (`setsid` when available, matching the process-session launcher) whose consumer cuts the stream into fixed-size parts with `dd bs=<part> count=1 iflag=fullblock`, publishes each with an atomic `mv`, then blocks until the host has read and removed it before producing the next. At most one part is ever on the guest disk and no archive is materialized. The host loop waits for the next part or the completion marker, reads it, deletes it (failing fast if the delete fails, since the producer would otherwise block), and feeds it to the existing `SnapshotUpload`. The size cap is enforced on the running total.

**Consumer failures are fatal, not silent.** `dd` buffers a whole part before its single write, so tar reaches EOF and exits 0 before the final part is on disk; a disk-full failure on that part is invisible to tar. The consumer therefore exits nonzero on any `dd` or `mv` failure and its status overrides tar's in the completion marker; `dd`'s diagnostics are kept per part and appended to the marker's stderr only on failure, so the host error reads e.g. `producer failed (exit 1): dd: ... No space left on device` and the upload is aborted. (Found by review; covered by a new test.)

**Scratch lives in one per-run directory, created before tar starts.** E2B and Modal keep the tar path inside the home. With parts appearing and disappearing next to it while tar archived `.`, GNU tar noticed the home root's mtime moving and exited 1 with `tar: .: file changed as we read it` (bsdtar has no such check, so this only showed on Linux CI, flakily, as "first snapshot on a fresh box failed"). Creating `<tarpath>.<runtag>/` up front and keeping list, parts, markers, stderr and pid inside it leaves `.` untouched during the archive. The directory is pruned from the archive by the existing in-home scratch prune. The per-run tag also means a producer orphaned by an aborted run (upload failure, timeout, or a kill that could not reach it) can never be mistaken for the current run's output; cleanup kills every recorded producer group and removes all scratch.

**Atomic promotion.** The existing S3 multipart upload already writes to an invisible upload and only replaces the object at `CompleteMultipartUpload`; an aborted upload leaves the previous object untouched. Rather than adding a temp key plus a server-side copy of up to 2 GB, the change keeps that mechanism and adds a guard so a run that produced no data can never complete. That guard makes the store's zero-parts branch (abort + write an empty object) unreachable, so it is removed.

**Rotation.** The AWS rotation keeps serving the stale body and reports `rotate_snapshot_failed` when the snapshot fails, and holds further rotation attempts on that scope for ten minutes (mirroring Modal's `rotationHoldUntil`) so an unhealthy store is not hammered with a full snapshot on every provision; a successful teardown snapshot lifts the hold. The body is terminated only after a successful snapshot. `reapDeepIdle` already skipped termination on snapshot failure and is unchanged.

Known limitation, unchanged from before this PR in effect: an alive body whose exec daemon is unreachable cannot be snapshotted and so cannot be rotated; it is retired when AWS enforces `maximumDurationInSeconds`, at which point the previous successful snapshot is restored.

## Tests

`test/home-snapshot.test.ts` (runs the real scripts against a temp dir):
- new: the tar is streamed with no archive and at most one part on the sandbox disk (inspects the sandbox tree at every part read)
- new: a guest write failure on the final part fails the snapshot instead of committing a truncated archive (a `dd` shim on `PATH` fails the last part with ENOSPC; asserts `complete()` never ran, the upload was aborted, the message is surfaced, and the previously stored snapshot is intact)
- new: a producer orphaned by a failed upload cannot poison the next snapshot
- updated: the size cap aborts the upload once crossed; a failed upload leaves the previously stored snapshot intact
- all existing round-trip, deadline, prune, symlink and hydrate tests pass unchanged

`test/aws-sandbox.test.ts`:
- new: a rotation whose snapshot fails keeps serving the old body (not terminated, error reported, later changes survive once the snapshot store recovers and rotation completes)

`test/support/fake-microvm.ts` emulates the new guest scripts. The E2B and Modal fakes already run the real scripts through `sh`, which is how the GNU tar interaction surfaced.

Local results (macOS): home-snapshot 22/22, aws-sandbox 20/20, modal-sandbox 56/56, e2b-sandbox 39/39; `npm run typecheck` and `npm run lint` clean. Linux (Docker `node:24-bookworm-slim`, Debian's GNU tar and dash): e2b + modal + home-snapshot looped 8 times with zero snapshot failures (previously 2 to 4 per run); the only failures in that image are the two `python3`-dependent "large command output" tests, which fail identically on `main` there. The `setsid` launch branch was exercised with a stub `setsid` on `PATH`; `dd iflag=fullblock` was verified byte-exact across pipe boundaries.


ckaznocha:
We tested this change against an AWS MicroVM deployment and found two additions that may strengthen the implementation.

**Rotation-hold lifecycle**

Retaining the current VM after a failed rotation snapshot is the right state-safety outcome. One cleanup detail: `rotationHoldUntil` is currently cleared after a successful teardown snapshot, but it should also be cleared after:

- A successful rotation snapshot.
- Successful explicit scope destruction.
- Both successful and already-terminated deep-idle reap paths.

Otherwise, scopes that experienced a failed rotation leave stale entries in the Core process after their lifecycle ends. We added behavioral tests that destroy or reap a held scope, recreate the same scope, and verify the replacement does not inherit the old hold.

**Optional direct-stream input**

The staged-part producer is a substantial improvement for backends that expose only command and file operations. For AWS MicroVMs, we obtained a stronger result by adding an optional `AsyncIterable<Uint8Array>` input to the shared snapshot helper.

Our authenticated MicroVM agent exposes a fixed-purpose endpoint that starts `tar` with an argument vector and streams stdout directly to Core. Core coalesces that response into the existing sequential S3 multipart upload. This writes neither a complete archive nor a temporary part to the guest filesystem.

The direct-stream path uses the same safety properties as this PR:

- One deadline covers stream acquisition, reads, uploads, and completion.
- The accumulated size is checked before uploading the offending part.
- Stream, timeout, upload, and size failures abort the multipart upload.
- The iterator is cancelled when the consumer exits early, allowing the endpoint to kill its tar child.
- `CompleteMultipartUpload` runs only after clean EOF.
- An aborted replacement leaves the previously committed S3 object intact.

The existing staged-part implementation remains useful as the fallback for E2B, Modal, and other backends without a raw response stream.

One detail in this PR already handles an issue we observed in production: because the size check happens on fixed-size parts, the accumulated value at failure is only a lower bound, not necessarily the archive’s final size. Changing the error to “home tar exceeds the cap” is more accurate than reporting that counter as the final tar size.

PR #1356, opened 2026-09-18
Add motion to popovers, dialogs, disclosures, and controls in the web UI

time-attack:
## What

Seven surfaces in the web UI that snapped between states now transition. Values follow Emil Kowalski's `animate` skill (the most-installed animation skill on skills.sh): transform and opacity only, strong ease-out, entries under 200ms, exits faster than entries, a reduced-motion block that keeps the opacity fades and drops movement. No dependency added; everything is CSS plus one `startViewTransition` call.

| # | Surface | Mechanism | Before / after (60 fps, before left, after right) |
|---|---|---|---|
| 1 | Menus, popovers, tooltips (`.menu-popover`, `.session-menu-popover`, `.slash-popover`, `.qm-tooltip`) | `@starting-style` scale 0.96 + 4px slide from the trigger side, 150ms | [menus.mp4](https://raw.githubusercontent.com/yc-software/qm/demo/web-animations-recordings/recordings/menus.mp4) |
| 2 | Dialogs (native `<dialog>` + div modals) | fade + scale in 200ms, native dialogs fade out 150ms via `display`/`overlay` allow-discrete; backdrop fades | [dialog.mp4](https://raw.githubusercontent.com/yc-software/qm/demo/web-animations-recordings/recordings/dialog.mp4) |
| 3 | Press feedback (`.btn`, `.icon-btn`, `.navrow`, `.user-pill`, send, stop) + send-button enable fade | `:active` scale 0.97, 160ms | [press.mp4](https://raw.githubusercontent.com/yc-software/qm/demo/web-animations-recordings/recordings/press.mp4) |
| 4 | New session row in the sidebar | `@starting-style` fade + 4px slide, 180ms | [newrow.mp4](https://raw.githubusercontent.com/yc-software/qm/demo/web-animations-recordings/recordings/newrow.mp4) |
| 5 | Disclosures: work fold, thinking, tool rows, project children, archived, background output | `::details-content` transition with `content-visibility` allow-discrete; `@starting-style` for the non-details ones | [disclosure.mp4](https://raw.githubusercontent.com/yc-software/qm/demo/web-animations-recordings/recordings/disclosure.mp4) |
| 6 | Theme switch | `document.startViewTransition` crossfade, 200ms | [theme.mp4](https://raw.githubusercontent.com/yc-software/qm/demo/web-animations-recordings/recordings/theme.mp4) |
| 7 | Copy confirmation check | 160ms pop from scale 0.6 | [copy.mp4](https://raw.githubusercontent.com/yc-software/qm/demo/web-animations-recordings/recordings/copy.mp4) |

Individual clips: `https://raw.githubusercontent.com/yc-software/qm/demo/web-animations-recordings/recordings/<name>-before.mp4` and `<name>-after.mp4` on branch `demo/web-animations-recordings`.

Also fixes the keychain spinner, which referenced an undefined `chat-spin` keyframe and never spun; it now uses the vendored `spin`.

## Why these seven

The skill gates motion by frequency: keyboard-driven and 100+/day actions stay instant. Everything here is occasional (menus, dialogs, disclosures, theme) or a sub-200ms feedback cue (press, copy). The sidebar collapse stays instant on purpose; `sidebar-rail-source.test.ts` locks that.

## Notes for review

- Closed `<details>` content is hidden with `content-visibility` in Chrome, so `@starting-style` on the body never fires; the `::details-content` transition is the working pattern. Browsers without it show content instantly.
- Session rows use keyed `repeat()`, so a row that moves gets re-inserted and fades once (180ms). Judged acceptable as state indication.
- Share dialog removes itself on close, so it gets the enter animation only.

## Verification

- `npm run typecheck` and `npm test` in `plugins/web-ui`: 1338 pass, 0 fail.
- Recorded headlessly against a web-only dev instance with CDP screencast; frame-diff analysis confirms multi-frame transitions in every after clip where the before clip has a single-frame jump.


time-attack:
Independent review pass (fresh-context agent) and how each finding landed in the second commit:

- **JS-placed popovers measured their own starting-style transform.** `placeMenuPopover`, `placeSessionMenu` and `placeLoadout` read `getBoundingClientRect` in the frame the popover is unhidden, so a scale/translate start skewed overflow and drop-up math. `.menu-popover` and `.session-menu-popover` now fade only; the slash popover and tooltip keep the slide. Fixed.
- **Reduced-motion rule shared a selector list with `::details-content`.** Engines without that pseudo-element would have dropped the whole rule. The `::details-content` selectors now sit in their own rule. Fixed.
- **Dialog exit half was dead code.** Every native `<dialog>` caller removes the element on close, so the `display`/`overlay` allow-discrete exit never ran. Removed; entry rules are shared with the div-based modals. Fixed.
- **Details reveal faded then snapped on auto-close.** Transition now lives on the `[open]` state only, so open animates and close is instant as before. `.stopped-work` joined the list. Fixed.
- **Spinner keyframe.** The reviewer flagged `spin` as undefined in `src/`; it is defined in the vendored `@earendil-works/pi-web-ui/app.css` that `shell.css` imports, confirmed via `CSSKeyframesRule` enumeration in the live page. `.mc-spinner` now uses it too and its private copy is gone.
- **Accepted as is:** a session row re-created by a `repeat()` key change (pin/unpin, rename swap) fades once and is a stacking context for 180ms; `.multi-select-color-popover`, `.kc-confirm` and `.sidebar-scrim` stay instant in this PR.

Menus, dialog and disclosure clips were re-recorded after the follow-up; links in the description point at the updated files. `npm run typecheck` and `npm test` in `plugins/web-ui`: 1338 pass, 0 fail.

PR #1357, opened 2026-09-18
Support files and images when steering running tasks

ReganBell:
Queued messages with files could not be steered because the live steering path only delivered text. This prepares attachments in the active sandbox, forwards native image input through all four harnesses, and enables Steer for attachment-only messages.

Moving a queued message into a signal is atomic and retry-safe. Live signals are acknowledged only after delivery; late signals remain available for terminal replay. Concurrent queue edits retain their updated message, preparation failures do not block Stop, and repeated clicks do not duplicate optimistic messages.

Validation: 160 affected core tests with real Postgres, 32 web tests, core/web typechecks, and lint passed. Two independent adversarial reviews plus a rebase integration review are resolved. Live Codex web QA covers text files, captionless images, and repeated Steer clicks; Firefox Slack smoke QA passed. Other harnesses have automated coverage.

Visual QA evidence is attached below. No mock backend was used.


ReganBell:
Final-version live image QA: a captionless screenshot was steered into an active Codex turn in a fresh conversation. The agent transcribed the tooltip and identified its background color correctly. Real Postgres and provider, no mocks.

clipboard

PR #1426, opened 2026-09-19
Link a person's sign-ins to one canonical principal

ReganBell:
## Problem

A founder who reaches qm through Slack, an email magic link, and a trusted OIDC provider gets three principal ids and is treated as three people. In production a secret drop minted from Slack belonged to the email principal, and the same founder signed in through OIDC was told the link was for someone else. Owner checks use `samePerson`, which only folds email case.

## Change

**Principal links.** An org admin records that a sign-in id belongs to a canonical principal, with evidence. Links live in the `principal_links` durable map of the company database and refresh with the identity service.

**Resolution at the layer every owner check flows through.** `personKey` and `samePerson` fold a linked sign-in to its canonical id, so ownership recorded under either id belongs to the person: keychain credentials, admin grants, approvals, secret drops, session participants, the admin users list.

**Canonical ids on new writes.** The identity service classifies a linked sign-in as its canonical principal, so Slack and web turns write the canonical id. The portal asks core for the canonical principal of the session subject (`GET /v1/principals/:id/canonical`, cached one minute) before proxying, and core canonicalizes the verified portal actor and the admin actor. The keychain saves credentials under the canonical owner.

**Rules.** The directory member stays canonical: the admin API refuses a link whose linked side is a directory member. Links are flat (no chains, a linked id cannot be canonical for others), require evidence, and are reversible. Like grants and impersonation they are portal-only; the agent cannot manage them. Deactivating either id deactivates the person.

## API

```
GET    /v1/admin/principal-links
POST   /v1/admin/principal-links      { principalId, canonicalId, evidence }
DELETE /v1/admin/principal-links/:principalId
GET    /v1/principals/:id/canonical   (source auth)
```

## Not in this change

Re-owning legacy rows (sessions, memory scopes, files) written under a linked sign-in before the link existed. Owner checks already honor them through `samePerson`; personal scopes (`personal:<id>`) stay split until a per-person migration moves them.

## Verification

Typecheck, lint, and the affected suites pass locally: principal-links (new), person-identity, identity, identity-offboarding, keychain, secret-drop, admin-keychain, admin-users, admin-external-users, admin-agent-capability, admin-scopes-directory, trusted-admin, portal-identity, portal-identity-gate, directory-store, directory-resolve, route-auth-conformance, agent-api-parity, agent-api-discovery, open-speaker-keychain, keychain-ask, sent-chat, loop-routes, slack-identity, plus the portal and admin plugin suites. Docs: `docs/principal-links.md`, portal README.


ReganBell:
Independent adversarial review (fresh context, high effort) ran against the first commit. Findings and their resolution in the second commit:

- **Directory guard missed Slack ids** (a roster row carries both ids, so linking a member's Slack id would merge two people): the guard now checks principal id and Slack id.
- **A link could confer org admin without a grant**: a sign-in holding an org admin grant can only be linked to a canonical principal that is already an admin.
- **Deactivation became irreversible once linked** (store keyed by the folded person at write time, re-keyed at read time): deactivations and external members are stored under the raw folded id and folded only in memory; directory-sync protection is evaluated at check time.
- **Credential id split from owner**: `credId` now derives from the canonical owner, so the same person saving the same service from either sign-in rotates one record.
- **Staleness after unlink**: the canonical lookup forces an identity refresh; the portal cache is one minute and documented.
- **Portal fell open to the raw subject when core could not answer**: the portal now returns 503 for proxied paths instead, and the impersonation self-check compares canonically.

Accepted as documented limitations: records written during a link keep the canonical owner after unlink (a wrong link is an incident to reconcile, and links are portal-only with evidence and audit); the guard runs at create time only; the resolver is process-global and populated by the identity refresh.

Shard 4 of the core tests is failing on timing-sensitive tests unrelated to this change (a different test each run, including on `main` at 52f65c46); `test/web-ui-sse.test.ts` passes five of five runs locally.

PR #1460, opened 2026-09-20
Make startup readiness and shutdown explicit

ckaznocha:
## Summary

Make application startup own config, identity, and MCP readiness instead of launching database work during `buildApp()` construction. Migration shutdown could otherwise close the PostgreSQL stores while those startup queries were still running.

Expose the existing per-owner close operations for session and metrics stores, and await shutdown releases individually so a failed close does not prevent later cleanup. Update server, worker, migration, and smoke entrypoints together.

This is the first, independently useful prerequisite for the AG-UI integration proposed in #1449. It introduces no AG-UI endpoint or application-turn behavior. Subsequent changes will be submitted separately after their prerequisites merge.

## Validation

- TypeScript typecheck and ESLint passed in an isolated checkout containing only this change.
- Focused MCP tests: 15 passed.
- Migration/startup tests passed against disposable PostgreSQL with no database-test skips.
- Independent review checked startup callers, shutdown ownership, and the outgoing diff; no blocking findings remained.


ckaznocha:
Any chance I can get some eyes on this? I have the full AG-UI protocol implement on my fork, I would love to start getting them merged.

https://github.com/ckaznocha/qm/branches/active

